PR #7, opened 2016-06-16
Add hot reloading with livereactload

kevinrobinson:
Adds hot reloading for development via https://github.com/milankinen/livereactload.

@david-rosales this is super helpful for UI prototyping, we can try it out this morning!  It adds some plumbing so that when you save a file in your editor, that new code gets injected into the browser within a second or two, without having to reload the page. 😄

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,5 +18,10 @@
     "pg": "^5.1.0"
   },
   "keywords": [],
-  "license": "MIT"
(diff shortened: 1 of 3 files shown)


kevinrobinson:
selfie

http://thehustle.co/assets/pooch02.jpg
image

PR #8, opened 2016-06-16
Generalize evidence endpoint API

kevinrobinson:
This will let us use this for a second type of evidence with the simple implementation of putting everything in the same table and only allowing JSON payloads.  It's directionally what we might want in a service.

This also upgrades to node 6.2.1 to get better ES6 support.


kevinrobinson:
selfie

image

PR #9, opened 2016-06-16
Add flow to build setup, fix a few bugs and add some more files

kevinrobinson:
This adds https://flowtype.org/ to the project, to be able to gradually add type checking to the project.  This doesn't have a watch command yet, but can be run locally and is also added to the project's test command so that type failure will break the build.

I added `@flow` to two files and caught two small bugs which are fixed now as well :)

Here's an example of making a typo on `version`, the key of an object being passed to a function in another module:
screen shot 2016-06-16 at 8 59 00 am

And the output:
screen shot 2016-06-16 at 8 59 06 am


kevinrobinson:
selfie

image

PR #14, opened 2016-06-16
Adding in student data

david-rosales:
I added in all the data from the list of students. I also decided to move the `allStudents` and `allQuestions` constants to a separate file so that the important code in `message_popup_page.jsx` is easily accessible without having to scroll down too much. I also had to adjust some of the prop settings in `student_card.jsx` to account for the fact that not all the students have every attribute.


david-rosales:
@kevinrobinson I got the initial student data in, can you review it?

ui/src/message_popup/data_lists.jsx
```diff
@@ -0,0 +1,142 @@
...
+    id: 8,
+    name: 'Ada',
+    grade: '7th grade',
+    gender: 'female',
+    race: 'Haitian',
+    behavior: null,
+    ell: 'Yes, Haitian Creole is native language',
+    learningDisabilities: 'Auditory disability',
+    academicPerformance: null,
+    interests: 'High-achiever, teacher\'s pet. Wants to be a doctor, is in yearbook',
```
kevinrobinson: In ES6 you can also use a backtick as a quote if you like, so then you don't have to use the backslash to escape:
```
`High-achiever, teacher's pet. Wants to be a doctor, is in yearbook`
```

ui/src/message_popup/data_lists.jsx
```diff
@@ -0,0 +1,142 @@
...
+    interests: 'Prefers ELA over math and science. Plays drums at home with his family. Has an after school job',
+    familyBackground: 'Younger sister',
+    ses: null
+  },
+  { 
+    id: 10, 
+    name: 'Sasha',
+    grade: '7th grade',
+    gender: 'female',
+    race: '?'
```
kevinrobinson: Let's give her a race so it feels authentic (given that there are all sorts of questions about presenting these profiles in a way that avoids people defaulting to stereotypes, and we'll have to improve that for sure).  I rolled on http://neilcic.com/identity.html and Sasha is now African American. :)

kevinrobinson:
@david-rosales This looks good!  I'm looking at why the [build is failing now](https://travis-ci.org/mit-teaching-systems-lab/threeflows/builds/138184227), sorry about that.  Either way I'll probably disable the typechecker tomorrow and then just work the bugs out with the typechecking before putting it in the build.

david-rosales:
@kevinrobinson sounds good, I'll merge it now.

kevinrobinson:
@david-rosales ah you won't be able to actually deploy your changes though - the failing build will prevent that (which is what we want in general).  I just pushed a commit fixing this, but you beat me to it, so I'll open a separate PR with that now :)

PR #15, opened 2016-06-16
Fix bug with partial propTypes interfering with Flow annotations

kevinrobinson:
Addressing https://github.com/mit-teaching-systems-lab/threeflows/pull/14#issuecomment-226600533

Not sure exactly why this is happening, some interplay between the Flow type annotations and the React PropTypes.  This also removes the Flow check from the `test` step - I'll debug this more on my box first.


kevinrobinson:
@david-rosales This should clear the build and then deploy after, and we can watch here! https://dashboard.heroku.com/apps/threeflows/activity

PR #17, opened 2016-06-16
Add states and toggles for Message PopUp practice modes

kevinrobinson:
This adds some state and toggles to the Message Pop Up game, configuring how much scaffolding is present for each popup.  Not all toggles work yet, so some are disabled.
screen shot 2016-06-16 at 6 21 11 pm

The mode state also gets threaded through as evidence (so we know what levels of support were present).
screen shot 2016-06-16 at 6 34 54 pm

### DIFF
--- a/ui/src/message_popup/message_popup_page.jsx
+++ b/ui/src/message_popup/message_popup_page.jsx
@@ -4,6 +4,8 @@ import React from 'react';
 import * as Routes from '../routes';
 import * as PropTypes from '../prop_types.js';
 import RaisedButton from 'material-ui/RaisedButton';
+import Toggle from 'material-ui/Toggle';
+import Divider from 'material-ui/Divider';
 import PopupQuestion from './popup_question.jsx';
 import type {Response} from './popup_question.jsx';
 import TextField from 'material-ui/TextField';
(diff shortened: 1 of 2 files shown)


kevinrobinson:
@david-rosales Can you review?  And also please suggest any design changes you think are worth trying to!

david-rosales:
@kevinrobinson It looks good! Just one thing I might change is that instead of a `showHintImmediately` type of check, I think it should be a `allowedToToggleHint` type of check. I think that when going through this, most if not all of the teacher's would like to try to come up with an idea before seeing an example. So that would mean that the example would be toggled off by default initially with a "need a hint?" button in the section where it would show. If they aren't allowed to see the hint, then they wouldn't have the option and the section would not be there. 

As for the styling, I definitely like the dividers a lot more than the difference in color and I think we should do that for the StudentCards as well!

Approved!

kevinrobinson:
@david-rosales This is awesome feedback, thanks! 😄 I'll make these changes tomorrow.

As for the color difference with student cards, I was thinking that color might be a consistent theme to signify "student card" across different games.  But that's just a thought about the future - I agree that the dividers would probably look better for this particular game so let's do it!

kevinrobinson:
@david-rosales This is the change from "do you want to see a model before answering every time" to "do you want to have hints available when you need them?"

screen shot 2016-06-17 at 9 10 16 am

PR #18, opened 2016-06-16
Add Babel transform for Object spread

kevinrobinson:
I guess this isn't supported in ES6 proper yet, but this is a super useful CoffeeScript-ism so let's use it!  http://redux.js.org/docs/recipes/UsingObjectSpreadOperator.html for more info.

It lets us change code like this:
```
    logFn(Object.assign(response, {
      name: this.state.name,
      clientTimestampMs: new Date().getTime()
    }));
```

to:
```
    logFn({
      ...response,
      name: this.state.name,
      clientTimestampMs: new Date().getTime()
    }));
```

This isn't so much for clarity as it is for consistency with how the spread works for arrays, and for how it works in JSX.

### DIFF
--- a/ui/.babelrc
+++ b/ui/.babelrc
@@ -4,6 +4,7 @@
     "development": {
       "plugins": [
         ["transform-flow-strip-types"],
+        ["transform-object-rest-spread"],
         ["react-transform", {
           "transforms": [{
             "transform": "livereactload/babel-transform",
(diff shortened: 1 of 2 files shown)


kevinrobinson:
@david-rosales Does this seem reasonable to you?

david-rosales:
@kevinrobinson Looks good!

PR #20, opened 2016-06-17
Added a hint card to message popup

david-rosales:
I added in the toggle hints functionality. The design layout is subject to change but just to keep the text area from being overwhelmed with examples whenever the hints were toggled I decided to display one at a time and have a button to cycle through them.

Below is how the popup looks before the hint is expanded.

This is the section after it is expanded.

### DIFF
--- a/ui/src/message_popup/data_lists.jsx
+++ b/ui/src/message_popup/data_lists.jsx
@@ -1,171 +1,248 @@
 export const allStudents = [
   { 
     id: 1,
-    name: 'Kevin',
-    grade: '7th grade',
-    gender: 'male',
-    race: 'Hispanic',
-    behavior: 'Disruptive in class',
+    name: `Kevin`,
+    grade: `7th grade`,
+    gender: `male`,
+    race: `Hispanic`,
+    behavior: `Disruptive in class`,
     ell: null,
     learningDisabilities: null,
-    academicPerformance: 'Grades are C/D and borderline failing. Failed science the last two semesters. Reading and writing is at the 5th grade level. Arrives late to class most of the time.',
+    academicPerformance: `Grades are C/D and borderline failing. Failed science the last two semesters. Reading and writing is at the 5th grade level. Arrives late to class most of the time.`,
     interests: null,
     familyBackground: null,
     ses: null,
   },
   {
     id: 2,
-    name: 'Floyd',
-    grade: '7th grade',
-    gender: 'male',
-    race: 'Caucasian',
-    behavior: 'Attendance issues',
+    name: `Floyd`,
+    grade: `7th grade`,
+    gender: `male`,
+    race: `Caucasian`,
+    behavior: `Attendance issues`,
     ell: null,
-    learningDisabilities: 'ADHD',
+    learningDisabilities: `ADHD`,
     academicPerformance: null,
-    interests: 'Wants to be a firefighter',
-    familyBackground: 'Divorced parents, younger brother',
-    ses: 'Free and reduced lunch'
+    interests: `Wants to be a firefighter`,
+    familyBackground: `Divorced parents, younger brother`,
+    ses: `Free and reduced lunch`
   },
   {
     id: 3,
-    name: 'Maia',
-    grade: '7th grade',
-    gender: 'female',
-    race: 'Caucasian',
-    behavior: 'Discipline report',
+    name: `Maia`,
+    grade: `7th grade`,
+    gender: `female`,
+    race: `Caucasian`,
+    behavior: `Discipline report`,
     ell: null,
     learningDisabilities: null,
     academicPerformance: null,
-    interests: 'In theater and loves art',
-    familyBackground: 'Single mom, two olders sisters',
+    interests: `In theater and loves art`,
+    familyBackground: `Single mom, two olders sisters`,
     ses: null
   },
   { 
     id: 4, 
-    name: 'Hayin',
-    grade: '7th grade',
-    gender: 'female',
-    race: 'Korean',
+    name: `Hayin`,
+    grade: `7th grade`,
+    gender: `female`,
+    race: `Korean`,
     behavior: null,
     ell: null,
     learningDisabilities: null,
     academicPerformance: null,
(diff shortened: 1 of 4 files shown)


david-rosales:
@kevinrobinson Can you review it? Also, should the titles of the example and non-example be good examples/bad examples like I have them or just examples/non-examples?

ui/src/message_popup/data_lists.jsx
```diff
@@ -1,171 +1,248 @@
 export const allStudents = [
   { 
     id: 1,
-    name: 'Kevin',
-    grade: '7th grade',
-    gender: 'male',
-    race: 'Hispanic',
-    behavior: 'Disruptive in class',
+    name: `Kevin`,
```
kevinrobinson: This is good for consistency, but I'd propose that if there's no escaping necessary we use single quotes.  In general, we want most static strings to be static and only have ones be dynamic or escaped if they need to.

I'm not sure of a better reason *why* but I think simplicity is the best one.  Also performance here won't matter but it's possible it's slightly more work for the JS engine to interpret these than not.  In general though I think if it's a static string we should use quotes and if it needs escaping or interpretation then we use the ticks.

ui/src/message_popup/hint_card.jsx
```diff
@@ -0,0 +1,141 @@
+// @flow
+import React from 'react';
+import * as PropTypes from '../prop_types.js';
+import RaisedButton from 'material-ui/RaisedButton';
+import Divider from 'material-ui/Divider';
+
+/*
+Shows a card for a student from the virtual school.
```
kevinrobinson: Stale comment here, can you update?  This would be a good place to put your comment in the PR about "this shows a hint and cycles through examples and non-examples"

ui/src/message_popup/hint_card.jsx
```diff
@@ -0,0 +1,141 @@
...
+      next = 0;
+    }
+    this.setState({nonExampleIndex: next});
+  },
+  
+  onHintsToggled(){
+    this.setState({ hidden: !this.state.hidden });
+  },
+
+  render() {
```
kevinrobinson: This looks awesome!  Here's a design suggestion.  I agree there's too much going on if we show all the examples and non-examples at once.  How about the button "Show hint" shows one hint, which is either an example or non-example.  And then the "Next hint" becomes "Another hint" and when that's click it shows another example or non-example.

Let me know if it's easier to chat in person too!

kevinrobinson: I think this also could help simplify the layout of the component here in code too let me know what you think.

ui/src/message_popup/hint_card.jsx
```diff
@@ -0,0 +1,141 @@
...
+    margin: 10,
+  },
+  prompt: {
+    fontSize: 16,
+    fontWeight: 'bold',
+    display: 'inline-block',
+  },
+  button: {
+    display: 'inline-block',
+    float: 'right'
```
kevinrobinson: In general, see if you can use [flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) for layout whenever possible over floats.  Flexbox is newer but well-supported now and is simpler to work with for more complex layouts than float.  For this it would be setting `display: 'flex'` on the `buttonRow` and setting an `alignItems` property on it to make the items align to a particular side.

We can talk this out in person too if that's easier :)

ui/src/message_popup/message_popup_page.jsx
```diff
@@ -147,8 +151,8 @@ export default React.createClass({
             <Toggle
               label="With hints available"
               labelPosition="right"
-              toggled={false}
-              disabled={true} />
+              toggled={this.state.allowedToToggleHint}
+              onToggle={this.onHintsOptionToggled} />
```
kevinrobinson: We'll have to pass `this.state.allowedToToggleHint` in to the `PopupQuestion` component too, won't we?  That way if we can use it [here](https://github.com/mit-teaching-systems-lab/threeflows/pull/20/files#diff-49aaf0f1ebf22819c85f8d7790ddbb18R85) and we can hide the "Show hint" button if it's false.

ui/src/message_popup/popup_question.jsx
```diff
@@ -104,6 +111,11 @@ export default React.createClass({
 
 
 const styles = {
+  hintCard: {
+    backgroundColor: '',
```
kevinrobinson: Do we need the empty string here?

kevinrobinson:
@david-rosales This looks great, awesome work on taking the idea and turning it into a product design, and then building it out!

My main feedback is on the design, and whether we can simplify with having the hints be a single example/non-example, and then "another hint" cycling through those.  So I'd suggest figuring out what you want to do there first.

david-rosales:
@kevinrobinson Thanks for the feedback! I'll start working on the fixes and I'll try out the single example design. I'll leave this PR open and commit to it again with the new changes.

kevinrobinson:
@david-rosales No problemo, sounds like a good plan! 👍

david-rosales:
@kevinrobinson I updated the PR with the changes based on your feedback. I also tried out the single example design and I think it looks great! Can you review it?

david-rosales:
@kevinrobinson I just realized that I never pushed my most recent changes, can you review it?

ui/src/message_popup/hint_card.jsx
```diff
@@ -49,7 +49,7 @@ export default React.createClass({
         {hidden &&
           (<div>
             <div style={styles.buttonRow}>
-              <div style={styles.prompt}>Need a hint?</div>
+              <div></div>
```
kevinrobinson: Can you remove this altogether if it's just a blank `div`?

kevinrobinson:
@david-rosales Looks good, approved!

ui/src/message_popup/hint_card.jsx
david-rosales: The blank div was there so that I could use the same style for the buttonRow with flex `justifyContent: 'space-between'` for both when the example is hidden and when it is shown.

PR #25, opened 2016-06-17
Adding eslint and doing some cleanup from it

kevinrobinson:
This adds http://eslint.org/ to the dev setup, using the recommended rules and React plugin as well as a few other rule changes that are from me trying it out.  This PR also contains some cleanup that came up from running it.

This isn't in the build process and won't fail the build or deploy, this is still just in the "trying it out locally" stage.


kevinrobinson:
This is most useful for catching typos and bugs like this (also a little useful for enforcing style conventions):
screen shot 2016-06-17 at 5 48 42 pm

screen shot 2016-06-17 at 5 52 04 pm

kevinrobinson:
fyi @david-rosales this might cause some merge conflicts too.  They all should be mechanical and small changes, but we can figure out next week.

david-rosales:
@kevinrobinson Looks good!

kevinrobinson:
Also brought upstream/master up to passing for both eslint and flow.

PR #26, opened 2016-06-20
Adding in the feedback feature

david-rosales:
I added in the preliminary feedback feature that should log things correctly (I could not test whether it would log correctly on heroku but it was working properly in `localStorage`). I also created the `RadioButtonGroup` on the introduction card of the message popup game. 

The feedback looks like this:

It currently just randomly picks a good example to show as sample feedback. However, in the future we will change it to be real examples by previous users.

### DIFF
--- a/ui/src/message_popup/feedback_card.jsx
+++ b/ui/src/message_popup/feedback_card.jsx
@@ -1,0 +1,114 @@
+import React from 'react';
+import _ from 'lodash';
+import * as PropTypes from '../prop_types.js';
+import RaisedButton from 'material-ui/RaisedButton';
+import TextField from 'material-ui/TextField';
+import Divider from 'material-ui/Divider';
+import TextChangeEvent from '../types/dom_types.js';
+
+/*
+This shows a feedback card after the user first presses save
+*/
+export default React.createClass({
+  displayName: 'FeedbackCard',
+
+  getInitialState: function(){
+    var example = _.shuffle(this.props.examples)[0];
+    return ({
+      finalResponseText: '',
+      example: example
+    });
+  },
+  
+  propTypes: {
+    initialResponseText: React.PropTypes.string.isRequired,
+    onDonePressed: React.PropTypes.func.isRequired,
+    examples: React.PropTypes.array.isRequired
+  },
+  
+  onTextChanged({target:{value}}:TextChangeEvent){
+    this.setState({ finalResponseText: value });
+  },
+  
+  onRevise(){
+    this.props.onDonePressed(true, this.state.finalResponseText);
+  },
+  
+  onPass(){
+    this.props.onDonePressed(false, '');
+  },
+
+  render() {
+    return (<div>
+      <div style={{margin: 10}}><Divider /></div>
+      <div style={styles.exampleBox}>
+        <div style={styles.exampleTitle}>Good Example</div>
+        <div style={styles.exampleText}>{this.state.example}</div>
+      </div>  
+      <div style={{margin: 10}}><Divider /></div>
+        
+      <div style={styles.textAreaContainer}>
+        <TextField
+          style={styles.textField}
+          textareaStyle={styles.textareaInner}
+          underlineShow={false}
+          floatingLabelText='Revise your response'
+          onChange={this.onTextChanged}
+          defaultValue={this.props.initialResponseText}
+          multiLine={true}
+          disabled={this.state.isRevising} 
+          rows={2}/>
+      </div>
+      <div style={styles.buttonRow}>
+        <RaisedButton
+          onTouchTap={this.onRevise}
+          style={styles.button}
+          primary={true}
+          label='Revise'/>
+        <RaisedButton
+          onTouchTap={this.onPass}
+          style={styles.button}
+          primary={true}
+          label='Pass'/>
+      </div>
+    </div>);
+  }
+});
+
+const styles = {
+  exampleBox: {
+    display: 'flex',
+    flexDirection: 'column',
+    margin: 20
+  },
+  exampleTitle: {
+    fontSize: 16,
+    fontWeight: 'bold',
+    marginTop: 6,
+    marginBottom: 15
+  },
+  exampleText: {
+    marginBottom: 10,
+    fontSize: 14
+  },
+  textAreaContainer: {
+    marginTop: 10,
+    margin: 20,
+    marginBottom: 10
+  },
+  textField: {
+    width: '100%'
+  },
+  textareaInner: {
+    border: '1px solid #eee',
+    marginBottom: 0
+  },
+  buttonRow: {
+    margin: 20,
+    marginTop: 10
+  },
+  button: {
+    display: 'inline-block',
+    marginRight: 10
+  },
+};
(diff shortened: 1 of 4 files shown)


david-rosales:
@kevinrobinson This is the implementation of the feedback help we talked about yesterday. I wasn't able to test it on an actual database so I'm not 100% sure that I did it correctly. However, it did work out with the `localStorage` so if it fails with the database then it should be small changes to get it working.
Can you review it?

ui/src/message_popup/message_popup_page.jsx
```diff
@@ -20,7 +21,7 @@ function randomizedQuestionsWithStudents() {
   });
 }
 
-function logLocalStorage(record) {
+function logLocalStorage(type, record) {
```
kevinrobinson: This doesn't look like it's using the `type` value anywhere here.  I can update that in a separate PR with the server-side changes too, I'll send that over next.

ui/src/message_popup/feedback_card.jsx
```diff
@@ -0,0 +1,117 @@
...
+import TextChangeEvent from '../types/dom_types.js';
+
+/*
+This shows a feedback card after the user first presses save
+*/
+export default React.createClass({
+  displayName: 'FeedbackCard',
+
+  getInitialState: function(){
+    var example = _.shuffle(this.props.examples)[0];
```
kevinrobinson: Nice use of passing a `prop` in and having that determine the initial state!

ui/src/message_popup/feedback_card.jsx
```diff
@@ -0,0 +1,117 @@
...
+    display: 'flex',
+    flexDirection: 'column',
+    margin: 20
+  },
+  exampleTitle: {
+    fontSize: 16,
+    fontWeight: 'bold',
+    marginTop: 6,
+    marginBottom: 2,
+    clear: 'right'
```
kevinrobinson: Do you need this here?  It doesn't seem like anything is using `float`.

ui/src/message_popup/message_popup_page.jsx
```diff
@@ -90,10 +95,16 @@ export default React.createClass({
     this.setState({ shouldShowStudentCards: !this.state.shouldShowStudentCards });
   },
   
-  onHintsOptionToggled(){
-    this.setState({ allowedToToggleHint: !this.state.allowedToToggleHint });
+  onHelpToggled(event, value){
```
kevinrobinson: This seems like it might be more natural to store the state as `scaffolding` and then the value can be `feedback | hints | none` rather than having two separate flags.  One benefit to that approach is that it makes it impossible to say `shouldShowFeedback: true, allowedToToggleHint: true`.

ui/src/message_popup/message_popup_page.jsx
```diff
@@ -172,16 +186,25 @@ export default React.createClass({
...
-            labelPosition="right"
-            toggled={false}
-            disabled={true}  />
-          <Toggle
-            label="With hints available"
-            labelPosition="right"
-            toggled={this.state.allowedToToggleHint}
-            onToggle={this.onHintsOptionToggled} />
+          <div style={{margin: 10}}><Divider /></div>
+          <RadioButtonGroup name="helpOptions" defaultSelected="none" onChange={this.onHelpToggled}>
```
kevinrobinson: Do you notice anything strange if you select a radio button, and then start typing in the text area?  I think you may need to change from using `defaultSelected` to using `valueSelected`.

React components for user input often have two ways to change their value, [controlled and controlled](https://facebook.github.io/react/docs/forms.html#controlled-components).

ui/src/message_popup/popup_question.jsx
```diff
@@ -54,25 +59,53 @@ export default React.createClass({
...
   onTextChanged({target:{value}}:TextChangeEvent) {
-    this.setState({ responseText: value })
+    this.setState({ initialResponseText: value })
   },
 
-  onSendPressed() {
-    const {elapsedMs, responseText} = this.state;
-    const {question, shouldShowStudentCard, allowedToToggleHint} = this.props;
+  onSavePressed() {
+    const {question, shouldShowStudentCard, allowedToToggleHint, shouldShowFeedback} = this.props;
```
kevinrobinson: Can you remove the values you don't need in this method?

ui/src/message_popup/popup_question.jsx
```diff
@@ -54,25 +59,53 @@ export default React.createClass({
...
       question,
       shouldShowStudentCard,
       allowedToToggleHint,
+      shouldShowFeedback,
       elapsedMs,
-      responseText
+      initialResponseText,
+      finalText
+    };
+    if(!this.state.isRevising){
```
kevinrobinson: I'd suggest moving this to separate methods, maybe `logResponse()` and `logRevision(finalText)` and `logRevisionDeclined`.  Those methods can then call the more general `logData` method if there are bits you want to re-use.  That way it's explicit when you call the method as to what will be logged, rather than having that be determined within `logData`.

kevinrobinson:
@david-rosales This is great work getting all the pieces connected together to make the feature work!  One small piece of design is feedback is that the horizontal indent for the example isn't there on the "hints" UI so it might seem more cohesive to have all the components flush to the same indentation on the left, up to you.

kevinrobinson:
@david-rosales From talking about this with Challenge 1.2 amigos, can we change the copy for the feedback to "Good example" instead of "Here is what someone else wrote"?

david-rosales:
@kevinrobinson I made the changes you suggested, can you review them?

Also, I can't remember how we resolved the merge conflict. Can you remind me what commands/steps we took so I can make note of them somewhere?

kevinrobinson:
@david-rosales Sure!  Let's do this together now if that's cool, and the GitHub docs are usually good for walkthroughs of how to do the steps: https://help.github.com/articles/resolving-a-merge-conflict-from-the-command-line/

ui/src/message_popup/message_popup_page.jsx
```diff
@@ -104,7 +108,9 @@ export default React.createClass({
       totalQuestions,
       questionsAnswered,
       shouldShowStudentCards,
-      allowedToToggleHint
+      helpType
+      //allowedToToggleHint,
+      //shouldShowFeedback
```
kevinrobinson: You can remove these.

ui/src/message_popup/popup_question.jsx
```diff
@@ -53,25 +56,68 @@ export default React.createClass({
...
+  
+  logRevision(finalText) {
+    this.logData(finalText);
+  },
+  
+  logRevisionDeclined(){
+    this.logData(undefined);
+  },
+  
+  logData(value) {
```
kevinrobinson: I'd suggest simplifying the branching here a bit more, and having `logData` take the type of message to record, and then optional params that are included in the response (eg., `finalText` for `logRevision`).

kevinrobinson:
@david-rosales Looks great, I'd update the `logData` method like we talked about and then approved!

david-rosales:
@kevinrobinson Alright, I added in the logData changes, want to review it one last time before the merge?

kevinrobinson:
@david-rosales sure, approved!  ship those logs 👍 
image

PR #27, opened 2016-06-20
Fix propType bug in VirtualSchoolPage, bring into Flow

kevinrobinson:
This was a typo bug, would have been caught by Flow so adding annotations so it will be next time. :)


kevinrobinson:
selfie, this is broken on the last deploy

PR #30, opened 2016-06-21
More copy tweaks on solution with Eliza and Jonathan's feedback

kevinrobinson:
screen shot 2016-06-21 at 11 00 00 am


kevinrobinson:
selfie

PR #32, opened 2016-06-21
Feature/velocity react animations

david-rosales:
Here I added in some velocity-react animations for message popup. They are definitely subject to change. I used solely used `VelocityTransitionGroup` components because I initially wanted to take advantage of the leave animations. However, a few of the animated components ended up not needing a leave animation and so might end up with a `VelocityComponent` instead if necessary.

### DIFF
--- a/ui/src/message_popup/hint_card.jsx
+++ b/ui/src/message_popup/hint_card.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import _ from 'lodash';
 import RaisedButton from 'material-ui/RaisedButton';
+import VelocityTransitionGroup from "velocity-react/velocity-transition-group";
 
 /*
 This shows a hint in the form of a toggleable good or bad example response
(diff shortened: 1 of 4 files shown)


david-rosales:
@kevinrobinson I added in the animations, can you review it?

ui/src/message_popup/hint_card.jsx
```diff
@@ -43,35 +44,39 @@ export default React.createClass({
...
-                style={styles.button}
-                secondary={true}
-                label="Show Example" />
-            </div>
-          </div>)
-        }
-        {!hidden &&
-          (<div>
-            <div style={styles.exampleBox}>
+        <VelocityTransitionGroup leave={{animation: "slideUp"}} runOnMount={true}>
```
kevinrobinson: This one felt a little clunky and seems like it might be more natural for it to `slideDown` from where the user clicks the button.  And am I following that the idea of the two different transition groups here is that the first time there's an animation but the second time there isn't, or it's different?  I'm wondering if it could be a single transition group with examples that `slideDown` when they enter and exit, would that make sense?

ui/src/message_popup/message_popup_page.jsx
```diff
@@ -205,16 +213,19 @@ export default React.createClass({
...
 const styles = {
   done: {
-    padding: 20
+    padding: 20,
+    width: 360
   },
   instructions: {
-    padding: 20
+    padding: 20,
+    width: 360
```
kevinrobinson: What's this about?

kevinrobinson:
@david-rosales I decided to do the merging to get this onto master for tomorrow morning, and did that over here: https://github.com/mit-teaching-systems-lab/threeflows/pull/37

kevinrobinson:
oops, didn't mean to close sorry!

kevinrobinson:
@david-rosales Do you mind reviewing https://github.com/mit-teaching-systems-lab/threeflows/pull/37 after-the-fact when you have a chance?  And then we can close this out too!

PR #33, opened 2016-06-21
Message Pop: Allow candidate to choose learning objectives in practice mode

kevinrobinson:
This also splits out some of the data files and breaks them into different files.  A few new Message Pop questions are added to test that the selection works.

Practice mode:
screen shot 2016-06-21 at 5 44 45 pm

Solution mode:
screen shot 2016-06-21 at 5 44 54 pm


kevinrobinson:
@david-rosales I'm going to merge this so we can use it tomorrow, but if you can do an after-the-fact that'd be awesome and I can incorporate any feedback you have Thursday.

PR #34, opened 2016-06-21
Add back superagent.end for logging

kevinrobinson:
`.send` doesn't send, `.end` ends the chain and makes the request.


kevinrobinson:
selfie

PR #35, opened 2016-06-21
Remove server dump from initial debugging

kevinrobinson:
This can be done through the console for now, no need to expose it.


kevinrobinson:
selfie

PR #40, opened 2016-06-22
Remove localStorage evidence, add broadcast to Slack

kevinrobinson:
This removes `localStorage` evidence collection from the UI, updates the server to work locally when there is no database and also adds an integration to broadcast evidence events in Slack (for observing testing).

Things should work the same locally, other than local storage and we should see new behavior in production after adding the Slack env variable.


kevinrobinson:
selfie, let's try it out during testing this morning

PR #41, opened 2016-06-22
Add session identifier for Message PopUp

kevinrobinson:
Also remove echoing the response from Postgres back to the client on logging evidence.


kevinrobinson:
selfie

PR #42, opened 2016-06-22
Revise button colors to change emphasis

kevinrobinson:
Revising the colors for the buttons.  @david-rosales I think we should revisit the Material UI theme at some point since this essentially swaps primary/secondary colors but seems like a helpful change for today to improve the relative salience of the different user actions.

screen shot 2016-06-22 at 8 57 03 am
screen shot 2016-06-22 at 8 57 18 am

screen shot 2016-06-22 at 8 57 29 am


kevinrobinson:
selfie-ing, adding a card around Material UI themes for after

PR #43, opened 2016-06-22
Add query endpoint back with basic auth

kevinrobinson:
Expose this for easy exploration and prototyping of scoring UIs, but put the tiniest barrier to access.  There's no real security yet since this is all friendly internal testing with nothing sensitive, but we'll likely add that for the next test.


kevinrobinson:
selfie

PR #44, opened 2016-06-22
Fix bug with returning result rows

kevinrobinson:
Bugfix from https://github.com/mit-teaching-systems-lab/threeflows/pull/43.


kevinrobinson:
selfie

PR #46, opened 2016-06-23
Fix bug in returning query rows

kevinrobinson:
Fixing a bug from last-minute changes before testing session.


kevinrobinson:
selfie

PR #49, opened 2016-06-24
Add hacked prototype for exploring responses (this is not the evaluation UI)

kevinrobinson:
This is hack-quality code and would need some revision before merging.  It's used to explore the responses from the playtest this week.  It introduces a few new dependencies (crossfilter, d3, dc) and the React interop is hacked together right now and that would need to be improved before merging as well.  This has only been used with hacks to read local data from disk.

Names are anonymized on the client, primarily just for sharing screenshots and could use a consistent hash so the anonymization is stable.

#### Logs and hovering to see questions and responses
screen shot 2016-06-23 at 8 28 06 pm

####  Screenshot of sessions and response logs on the left, and breakdowns and dimensions on the right:
screen shot 2016-06-23 at 8 22 30 pm

#### Zoomed-out, showing cross-selection and filtering across components:
screen shot 2016-06-23 at 8 22 25 pm


kevinrobinson:
@david-rosales I'll come back to this next week, it's a bit of a mess so need to review yet, just sharing to get it off my computer and into GitHub :)

david-rosales:
@kevinrobinson Looks really good so far! Approved

kevinrobinson:
@david-rosales I did some small cleanup, but I don't think it's worth it to improve the React interop now.  I had been thinking previously that we'd use this as part of an overview navigation for the scoring UI, but having shifted to try the mobile web UX, repurposing this makes less sense. :)

I'm going to merge this and keep it at prototype quality for now - we can make improvements to this separately as we need it, and it's in a separate route (now called `exploration`) so it's isolated from the work on the `scoring` UX.

PR #50, opened 2016-06-24
Adding in improvements based on the feedback

david-rosales:
I added in plenty of new features/fixes based on the playtest feedback from Wednesday/Thursday. I'll go through the visible ones with images below:

Changed button name from Save/Send to Respond/Save Response

Added a progress bar

Added a slider in the practice mode options page for the number of questions in the session

Summary after each question option as a breather for practice mode

There is now a final response time summary at the end of each session

PR #51, opened 2016-06-27
Remove data_lists, which was factored out

kevinrobinson:
I noticed this in https://github.com/mit-teaching-systems-lab/threeflows/pull/50/files, this was factored out previously but this file was still leftover.


kevinrobinson:
@david-rosales selfie, this is cleanup removing a dead file

PR #53, opened 2016-06-28
Refactoring the Instructions and Scaffolds to separate files

david-rosales:
I moved the Instructions and option scaffolding to separate components in separate files. Also, I started on making the practice page look less cluttered. This can be done by using query strings to designate which options should be shown. Having no query string will not show any options at the moment but all the default options are still selected. The purpose of this was so that we can discuss and decide on which ones should be always shown. 

For now, I have an example of that with the query string `basic` which shows the following options:

I experimented with using sliders on the side of the learning objectives for a little but I thought it looked a bit cluttered so I will experiment more with it tomorrow and see if I can make it look better with that.

### DIFF
--- a/ui/src/message_popup/experience_page.jsx
+++ b/ui/src/message_popup/experience_page.jsx
@@ -21,7 +21,6 @@ import InstructionsCard from './instructions_card.jsx';
 
 const ALL_COMPETENCY_GROUPS = 'ALL_COMPETENCY_GROUPS';
 
-
 /*
 Shows the MessagePopup game
 */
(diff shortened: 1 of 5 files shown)


david-rosales:
@kevinrobinson I started moving the different render methods from inside `experience_page.jsx` to separate component files. I ended up passing a lot of props to the `InstructionCard` so that it could just run in general. One thing I can do tomorrow is start moving some of the methods into both the new `instruction_card.jsx` and `scaffolding_card.jsx` files. Can you review it and give any suggestions on what would be best to keep in the `experience_page.jsx` and would would be best to move to the new files?

ui/src/message_popup/experience_page.jsx
```diff
@@ -237,102 +231,30 @@ export default React.createClass({
...
-              label="With no help available"
-              style={styles.radioButton}
-              />
-          </RadioButtonGroup>
-        </div>
-        <Divider />
-      </div>
-    );
+      <InstructionsCard 
+        isSolutionMode={isSolutionMode}
```
kevinrobinson: @david-rosales This is a great first step of refactoring, to pull this out and simplify the experience page!  There's a lot of parameters being passed in, so another step of simplification could be to not have the instructions/configuration pass all of the state back up (eg., maybe it's just "config" and "questions").

kevinrobinson:
@david-rosales I merged so that caused the conflict here, let me know if if you want a hand resolving!

PR #56, opened 2016-06-29
Add in Flow weak mode triggers for all files, fix up issues

kevinrobinson:
This adds in comments to trigger Flow in "weak mode" for all JS files, and does a little related cleanup.  It doesn't seem like there's a good way to control this at a finer granularity ([docs](https://flowtype.org/docs/existing.html#weak-mode)), but this at least runs every file through the checker.  It doesn't seem to catch much in the process, but is a good incremental step forward.


ui/src/components/side_menu.jsx
```diff
@@ -1,107 +0,0 @@
-import React from 'react';
```
kevinrobinson: This was leftover from an earlier prototype and we're not using it anymore so ✂️

kevinrobinson:
@david-rosales Can you review this when you get a chance?  It's a small change but in the right direction :)

david-rosales:
@kevinrobinson Looks good! Approved

PR #58, opened 2016-07-01
Adding auth0 for UI-side authentication, no server checks yet

kevinrobinson:
This uses the https://auth0.com/ service for authenticating the user's email address through Google.  This is mostly handled in the `AuthContainer` component, which then provides authorization information to child components via React context.

This is used to pre-fill the MessagePopUp experience with the user's name, and to record the email address of anyone scoring responses.

This is only enforced in the browser; although it does send a redirect through the server the server isn't looking at, guarding or checking any of this information (the existing basic auth is still there).  The auth0 setup here is using tokens that are client-side, so experimenting with passing them through and validating on the server that would be a next step (or switching directions to capture the OAuth tokens on the server and store them in a session).

### DIFF
--- a/ui/.flowconfig
+++ b/ui/.flowconfig
@@ -3,3 +3,4 @@
 .*/node_modules/config-chain/test/broken.json
 .*/node_modules/npmconf/test/.*
 .*/node_modules/react-motion/.*
+.*/node_modules/reqwest/.*
--- a/ui/package.json
+++ b/ui/package.json
@@ -20,6 +20,7 @@
   },
   "homepage": "https://github.com/mit-teaching-systems-lab/threeflows#readme",
   "dependencies": {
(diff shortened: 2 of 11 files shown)


kevinrobinson:
screen shot 2016-06-30 at 6 14 26 pm

screen shot 2016-06-30 at 8 48 56 pm

screen shot 2016-06-30 at 7 17 10 pm

kevinrobinson:
There's also a bug of some sort here with auth0 and md5 that I can't quite figure out.  Removing livereactload works around for now, so depending on how smoothly server guarding works might need to investigate that more.

kevinrobinson:
@david-rosales Can you review this at some point?

david-rosales:
@kevinrobinson Approved!

PR #59, opened 2016-07-01
Redirect to HTTPS in production

kevinrobinson:
No mas http.


kevinrobinson:
@david-rosales I'm going to merge this now, since the auth0 config is set to only allow interacting with https in production, and I'm trying to test how the auth bounce works on the phone.

Can you review post-merge?

david-rosales:
@kevinrobinson Looks good!

PR #60, opened 2016-07-01
Rename to eslintrc, so that Sublime plugin picks it up

kevinrobinson:
This came up in adding https://github.com/roadhump/SublimeLinter-eslint locally.


david-rosales:
@kevinrobinson Do we also need to update the npm command `lint-quiet` with the new name in `package.json`?

kevinrobinson:
@david-rosales great catch, fixed! 😄  I also noticed in the process that it was only running on `.jsx` files and not `.js`, so updated that and fixed the few things that revealed as well.

PR #62, opened 2016-07-06
Remove auth0 bounce, use self-identifying email form for playtest

kevinrobinson:
For the upcoming playtest, this removes the auth0 bounce to Google and instead just asks users to enter their email address.  There's no authentication of the address and there's no server-side checks.  It's stored in localStorage until the user "logs out" to clear it.

The experience page and scoring page are updated to use this email address.

#### Asking for email address
screen shot 2016-07-06 at 11 08 24 am

#### Updates to the experience page
screen shot 2016-07-06 at 11 10 00 am

### DIFF
--- a/ui/src/auth_container.jsx
+++ b/ui/src/auth_container.jsx
@@ -25,7 +25,7 @@ export default React.createClass({
 
   getDefaultProps() {
     return {
-      localStorageKey: 'threeflow_email_registration'
+      localStorageKey: 'threeflows_email_registration'
     };
   },
 
--- a/ui/src/auth_container.jsx
+++ b/ui/src/auth_container.jsx
@@ -1,12 +1,20 @@
 /* @flow weak */
+import _ from 'lodash';
 import React from 'react';
-import Auth0Lock from 'auth0-lock';
-import Auth0Variables from './auth0_config.js';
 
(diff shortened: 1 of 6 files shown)


kevinrobinson:
@david-rosales Can you review this?  This removes auth0 and uses a simpler self-report email address with no authentication.  The idea is that adding an authentication check via email for people who likely will not be on their own computers is more trouble than it's worth, especially with in-person observation.

ui/src/auth_container.jsx
```diff
@@ -17,117 +25,141 @@ export default React.createClass({
 
   getDefaultProps() {
     return {
-      localStorageKey: 'threeflows_auth_container'
+      localStorageKey: 'threeflow_email_registration'
```
david-rosales: This probably doesn't matter but maybe for consistency and so that it doesn't cause problems later, can we add the 's' to the 'threeflow' in the `localStorageKey` so that it will be `threeflows_email_registration` instead of `threeflow_email_registration`.

david-rosales:
@kevinrobinson It all looks good, i just made one minor suggestion that is probably still fine if not done. Approved!

ui/src/auth_container.jsx
kevinrobinson: Good catch!  Fixin'

PR #63, opened 2016-07-06
Revise Slack integration to remove any data

kevinrobinson:
This was useful for internal testing and monitoring, but doesn't make sense for a real playtest.  Separately I'll update this to a private channel.


kevinrobinson:
@david-rosales Does this make sense?  If we're doing this with outside people we don't really want to be putting any data from that in a channel that everyone on the team has access to anymore. :)

david-rosales:
@kevinrobinson Looks good

PR #64, opened 2016-07-06
Fix colors import case bug

kevinrobinson:
Trying to address the build failure (which doesn't happen locally), and removing extra auth0 dependency.


kevinrobinson:
selfie, trying to unblock the build.  cc @david-rosales

PR #67, opened 2016-07-06
Finishing up the refactoring of InstructionsCard

david-rosales:
I moved most of the functions from inside `instructions_card.jsx` into `scaffolding_card.jsx`. I decided to leave a lot of the configuration as a state inside of the instructions card because I couldn't come up with a good idea of when to pass the state from within `scaffolding_card.jsx` to `instructions_card.jsx` as done with `instructions_card.jsx` to `experience_page.jsx`. Because the parameters from `scaffolding_card.jsx` are changed and instantly updated, whereas the configuration from `instructions_card.jsx` is only saved when the start button is pressed, `instructions_card.jsx`'s state gets changed with the new parameters from within `scaffolding_card.jsx` every time something gets updated or changed inside `scaffolding_card.jsx`.

### DIFF
--- a/ui/src/message_popup/experience_page.jsx
+++ b/ui/src/message_popup/experience_page.jsx
@@ -51,7 +51,7 @@ export default React.createClass({
     const sessionId = uuid.v4();
     const hasStarted = false;
     const questionsAnswered = 0;
-    const responseTimes = []
+    const responseTimes = [];
     const gameSession = {sessionId, hasStarted, questionsAnswered, responseTimes};
     return {
       scaffolding,
@@ -178,9 +178,7 @@ export default React.createClass({
           <PopupQuestion
             key={JSON.stringify(question)}
             question={question}
-            shouldShowStudentCard={scaffolding.shouldShowStudentCard}
-            shouldShowSummary={scaffolding.shouldShowSummary}
-            helpType={scaffolding.helpType}
+            scaffolding={scaffolding}
             limitMs={this.state.limitMs}
             onLog={this.onLog}
             onDone={this.onQuestionDone}
--- a/ui/src/message_popup/experience_page.jsx
+++ b/ui/src/message_popup/experience_page.jsx
@@ -110,10 +110,9 @@ export default React.createClass({
   },
 
   render() {
-    const {
-      gameSession,
-      scaffolding
-    } = this.state;
+    const {gameSession, scaffolding} = this.state;
+    const {hasStarted, questionsAnswered} = gameSession;
+    const {sessionLength} = scaffolding;
     if (_.has(this.props.query, 'mobilePrototype')) return this.renderMobilePrototype();
     return (
       <div>
@@ -126,9 +125,9 @@ export default React.createClass({
               primaryText="Restart game" />
           }
         />
-        {!gameSession.hasStarted && this.renderInstructions()}
-        {gameSession.questionsAnswered >= scaffolding.sessionLength && this.renderDone()}
-        {gameSession.hasStarted && gameSession.questionsAnswered < scaffolding.sessionLength && this.renderPopupQuestion()}
+        {!hasStarted && this.renderInstructions()}
+        {questionsAnswered >= sessionLength && this.renderDone()}
+        {hasStarted && questionsAnswered < sessionLength && this.renderPopupQuestion()}
       </div>
     );
   },
@@ -155,25 +154,26 @@ export default React.createClass({
   },
 
   renderInstructions() {
+    const {scaffolding} = this.state;
+    const {sessionLength, email, helpType} = scaffolding;
     return (
       <InstructionsCard 
-        sessionLength={this.state.scaffolding.sessionLength}
+        sessionLength={sessionLength}
         onStartPressed={this.onStartPressed}
-        email={this.state.scaffolding.email}
+        email={email}
         itemsToShow={this.props.query}
-        helpType={this.state.scaffolding.helpType}
+        helpType={helpType}
         />);
   },
   
   renderPopupQuestion() {
-    const {
-      scaffolding,
-      gameSession
-    } = this.state;
-    const question = scaffolding.questions[gameSession.questionsAnswered];
+    const {scaffolding, gameSession} = this.state;
+    const {questionsAnswered} = gameSession;
+    const {questions, sessionLength} = scaffolding;
+    const question = questions[questionsAnswered];
     return (
       <div style={styles.container}>        
-        <LinearProgress color="#EC407A" mode="determinate" value={gameSession.questionsAnswered} max={scaffolding.sessionLength} />
+        <LinearProgress color="#EC407A" mode="determinate" value={questionsAnswered} max={sessionLength} />
         <VelocityTransitionGroup enter={{animation: "slideDown"}} leave={{animation: "slideUp"}} runOnMount={true}>
           <PopupQuestion
             key={JSON.stringify(question)}
@@ -182,7 +182,7 @@ export default React.createClass({
             limitMs={this.state.limitMs}
             onLog={this.onLog}
             onDone={this.onQuestionDone}
-            isLastQuestion={gameSession.questionsAnswered+1===scaffolding.sessionLength ? true : false}/>
+            isLastQuestion={questionsAnswered+1===sessionLength ? true : false}/>
         </VelocityTransitionGroup>
         <Snackbar
           open={this.state.toastRevision}
--- a/ui/src/message_popup/experience_page.jsx
+++ b/ui/src/message_popup/experience_page.jsx
@@ -111,8 +111,8 @@ export default React.createClass({
   render() {
     const {gameSession} = this.state;
     const hasStarted = gameSession !== undefined;
-    const questionsAnswered = hasStarted ? gameSession.questionsAnswered : undefined;
-    const questions = hasStarted ? gameSession.questions : undefined;
+    const questionsAnswered = hasStarted ? gameSession.questionsAnswered : 0;
+    const questions = hasStarted ? gameSession.questions : [];
     const sessionLength = hasStarted ? questions.length : 0;
     if (_.has(this.props.query, 'mobilePrototype')) return this.renderMobilePrototype();
     return (
@@ -127,7 +127,7 @@ export default React.createClass({
           }
         />
         {!hasStarted && this.renderInstructions()}
-        {questionsAnswered >= sessionLength && this.renderDone()}
+        {hasStarted && questionsAnswered >= sessionLength && this.renderDone()}
         {hasStarted && questionsAnswered < sessionLength && this.renderPopupQuestion()}
       </div>
     );
--- a/ui/src/message_popup/experience_page.jsx
+++ b/ui/src/message_popup/experience_page.jsx
@@ -47,21 +47,24 @@ export default React.createClass({
     const shouldShowSummary = !isSolutionMode;
     const sessionLength = 10;
     const scaffolding = {email, questions, shouldShowStudentCard, shouldShowSummary, helpType, sessionLength};
+
+    const sessionId = uuid.v4();
+    const hasStarted = false;
+    const questionsAnswered = 0;
+    const responseTimes = []
+    const gameSession = {sessionId, hasStarted, questionsAnswered, responseTimes};
     return {
       scaffolding,
-      sessionId: uuid.v4(),
-      hasStarted: false,
-      questionsAnswered: 0,
+      gameSession,
       toastRevision: false,
       limitMs: 90000,
-      responseTimes: []
     };
   },
 
   onStartPressed(scaffolding) {
     this.setState({
       scaffolding,
-      hasStarted: true
+      gameSession: _.set(_.clone(this.state.gameSession), 'hasStarted', true)
     });
   },
   
@@ -76,14 +79,16 @@ export default React.createClass({
   },
   
   addResponseTime(time){
-    this.setState({responseTimes: this.state.responseTimes.concat(time)});
+    var gameSession = _.clone(this.state.gameSession);
+    gameSession.responseTimes.push(time);
+    this.setState({gameSession});
   },
 
   onLog(type, response:Response) {
     Api.logEvidence(type, {
       ...response,
       name: this.state.scaffolding.email,
-      sessionId: this.state.sessionId,
+      sessionId: this.state.gameSession.sessionId,
       clientTimestampMs: new Date().getTime()
     });
   },
@@ -91,7 +96,9 @@ export default React.createClass({
   onQuestionDone(elapsedSeconds) {
     this.playToast();
     this.addResponseTime(elapsedSeconds);
-    this.setState({ questionsAnswered: this.state.questionsAnswered + 1 });
+    var gameSession = _.clone(this.state.gameSession);
+    gameSession.questionsAnswered += 1;
+    this.setState({ gameSession });
   },
 
   onDonePressed() {
@@ -104,13 +111,10 @@ export default React.createClass({
 
   render() {
     const {
-      hasStarted,
-      questionsAnswered,
+      gameSession,
       scaffolding
     } = this.state;
-    
     if (_.has(this.props.query, 'mobilePrototype')) return this.renderMobilePrototype();
-    
     return (
       <div>
         <NavigationAppBar
@@ -122,9 +126,9 @@ export default React.createClass({
               primaryText="Restart game" />
           }
         />
-        {!hasStarted && this.renderInstructions()}
-        {questionsAnswered >= scaffolding.sessionLength && this.renderDone()}
-        {hasStarted && questionsAnswered < scaffolding.sessionLength && this.renderPopupQuestion()}
+        {!gameSession.hasStarted && this.renderInstructions()}
+        {gameSession.questionsAnswered >= scaffolding.sessionLength && this.renderDone()}
+        {gameSession.hasStarted && gameSession.questionsAnswered < scaffolding.sessionLength && this.renderPopupQuestion()}
       </div>
     );
   },
@@ -137,7 +141,7 @@ export default React.createClass({
             <div style={styles.doneTitle}>You finished!</div>
             <Divider />
             <FinalSummaryCard 
-              responseTimes={this.state.responseTimes} 
+              responseTimes={this.state.gameSession.responseTimes} 
               limitMs={this.state.limitMs} />
             <RaisedButton
               onTouchTap={this.onDonePressed}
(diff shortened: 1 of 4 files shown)


david-rosales:
@kevinrobinson I did some more refactoring, can you review it?

kevinrobinson:
@david-rosales Ah, I think we miscommunicated a bit, although this looks like a good improvement too  :)

When we were talking about simplifying how the props are passed, I meant between `experience_page` and `instructions_page`.  The [experience_page#L47](https://github.com/mit-teaching-systems-lab/threeflows/blob/master/ui/src/message_popup/experience_page.jsx#L47) has a list of properties, and you've factored the `instructions_card` nicely so that they all are set as a unit in [onStartPressed](https://github.com/mit-teaching-systems-lab/threeflows/blob/master/ui/src/message_popup/experience_page.jsx#L71).  So my suggestion is to update `experience_page` so that `getInitialState` contains something like `scaffolding` and `gameSession`, and then `onStartPressed` can take just those two higher-level objects and replace them all at once in `onStartPressed`.

That will also allow removing `competencyGroupValue` from the `experience_page` state, and moving the `sessionId` into `onStartPressed` to clarify that that's when the new session starts.

Let me know if that makes sense, and if it's easier to chat about in Slack or in person let's do it!

david-rosales:
@kevinrobinson Ah, I see. That would make the handling of props much simpler. I do want to merge this before #83 though and after open another pull request for refactoring because similar changes will be necessary for the `MobileInterface` in #83.

kevinrobinson:
@david-rosales Sounds like a good plan! 👍

david-rosales:
@kevinrobinson I've made some more improvements and now there are `scaffolding` and `gameSession` states. The `scaffolding` state contains most of the props that were passed into `InstructionsCard` and the `gameSession` state contains some of the extra and top level info and doesn't get passed into any props. Can you review it when you get a chance?

ui/src/message_popup/experience_page.jsx
```diff
@@ -46,42 +44,35 @@ export default React.createClass({
...
-      isSolutionMode,
-      sessionId,
-      competencyGroupValue: ALL_COMPETENCY_GROUPS,
-      questions: allQuestions,
-      email: this.context.auth.userProfile.email,
-      hasStarted: false,
-      questionsAnswered: 0,
-      sessionLength: 10,
+      scaffolding: {
+        email: this.context.auth.userProfile.email,
```
kevinrobinson: Some of these seem like they're logically part of a session, rather than scaffolding (eg., `email`, `question` and `sessionLength`.  Does it make sense to move them there?

ui/src/message_popup/experience_page.jsx
```diff
@@ -46,42 +44,35 @@ export default React.createClass({
...
+        shouldShowStudentCard: isSolutionMode ? _.has(this.props.query, 'cards') : true,
+        shouldShowSummary: !isSolutionMode,
+        sessionLength: 10
+      },
+      gameSession: {
+        sessionId: uuid.v4(),
+        hasStarted: false,
+        questionsAnswered: 0,
+        responseTimes: []
+      },
```
kevinrobinson: We probably want to start with this being null - there's no `gameSession`to start, and then when the session is started it gets a new `uuid`.

ui/src/message_popup/experience_page.jsx
```diff
@@ -46,42 +44,35 @@ export default React.createClass({
...
     this.setState({
-      email,
-      sessionLength,
-      questions,
-      shouldShowStudentCard,
-      shouldShowSummary,
-      helpType,
-      hasStarted: true
+      scaffolding,
+      gameSession: _.set(_.clone(this.state.gameSession), 'hasStarted', true)
```
kevinrobinson: Using the ES6 spread operator might be more direct here, up to you:

```
const {gameSession} = this.state;
this.setState({
  scaffolding,
  gameSession: {...gameSession, hasStarted: true}
});

File: ui/src/message_popup/experience_page.jsx [hunk 4]
```diff
@@ -91,22 +82,26 @@ export default React.createClass({
...
     });
   },
   
   onQuestionDone(elapsedSeconds) {
     this.playToast();
     this.addResponseTime(elapsedSeconds);
-    this.setState({ questionsAnswered: this.state.questionsAnswered + 1 });
+    var gameSession = _.clone(this.state.gameSession);
+    gameSession.questionsAnswered += 1;
+    this.setState({ gameSession });
```
kevinrobinson: Similar comment here about using the ... spread operator.

ui/src/message_popup/experience_page.jsx
```diff
@@ -91,22 +82,26 @@ export default React.createClass({
   },
   
   addResponseTime(time){
-    this.setState({responseTimes: this.state.responseTimes.concat(time)});
+    var gameSession = _.clone(this.state.gameSession);
+    gameSession.responseTimes.push(time);
+    this.setState({gameSession});
```
kevinrobinson: ... spread for all the things! :)

ui/src/message_popup/popup_question.jsx
```diff
@@ -29,9 +29,9 @@ export default React.createClass({
   mixins: [SetIntervalMixin],
 
   propTypes: {
-    shouldShowStudentCard: React.PropTypes.bool.isRequired,
-    shouldShowSummary: React.PropTypes.bool.isRequired,
-    helpType: React.PropTypes.string.isRequired,
+    scaffolding: React.PropTypes.object.isRequired,
```
kevinrobinson: You can also use `React.PropTypes.shape` to express that scaffolding needs to have a few fields (and to have Flow check that those are the only ones we're accessing).

ui/src/message_popup/instructions_card.jsx
```diff
@@ -26,50 +26,39 @@ export default React.createClass({
...
-      shouldShowSummary: true,
-      helpType: this.props.helpType
+      scaffolding: { 
+        email: this.props.email,
+        sessionLength: this.props.sessionLength,
+        questions: allQuestions,
+        shouldShowStudentCard: true,
+        shouldShowSummary: true,
+        helpType: this.props.helpType
+      }
```
kevinrobinson: The way the state and props flow now, this component has to have most of the same bits as the experience component above.    That was my original idea, but perhaps in the actual execution it's more complicated so let's try sketching it out and see if there's a way to simplify.  I can swing by later this afternoon or we could do it in the morning tomorrow.

david-rosales:
@kevinrobinson I made the changes we discussed yesterday about moving the `ScaffoldingCard` into the `MessagePopupExperiencePage` rather than inside `InstructionsCard`. I also decided to have the 'Start' button and the email text field moved into the `ScaffoldingCard` so that we could have an internal state within it that gets adjusted and only goes back up to the top state when ready to create a session. As a result, the gameSession contains solely the email until a session actually starts (when `onSaveScaffoldingAndSession` is called). `InstructionsCard` also now only contains text and there isn't much going on inside it. I also moved some of the parts of scaffolding to gameSession since, as you pointed out, they make more sense to be there. Can you review the changes when you get a chance?

kevinrobinson:
@david-rosales This is awesome work and great progress!  There's a lot of pieces in this refactoring and you've made the flow of state between components a lot more direct. :thumbsup:

I left comments inline with more explanation, but here's the few steps I think we have left to finish this off!

##### Remove state that isn't needed
- [ ] `email` from experience_page state
- [ ] `hasStarted` from experience_page state
- [ ] `questions` from scaffolding_card state (it can be computed from other state)

##### Clarify how a gameSession gets created
- [ ] update `onSaveScaffoldingAndSession` so it creates a new `gameSession` from the pieces it's passed
- [ ] update `onSave` so it only passes scaffodling, email and questions

##### Cleanup
- [ ] Remove methods from `scaffolding_card` that were extracted out

ui/src/message_popup/experience_page.jsx
```diff
@@ -117,24 +98,32 @@ export default React.createClass({
     this.setState(this.getInitialState());
   },
 
+  onSaveScaffoldingAndSession(scaffolding, gameSession){
```
kevinrobinson: I think this second parameter isn't quite a full gameSession (ie., it doesn't have the `responseTimes` and the `sessionId` yet.  This method could take the scaffolding and then the email and set of questions, and it's job is to create a `gameSession` from those pieces and the other pieces needed.

ui/src/message_popup/experience_page.jsx
```diff
@@ -117,24 +98,32 @@ export default React.createClass({
...
+    this.setState({
+      scaffolding: {
```
kevinrobinson: You can simplify the syntax a bit, since
```
this.setState({
  scaffolding: {...scaffolding}
})
```

is equivalent to 

```
this.setState({
  scaffolding
});
```

ui/src/message_popup/experience_page.jsx
```diff
@@ -117,24 +98,32 @@ export default React.createClass({
...
+        ...scaffolding
+      },
+      gameSession: {
+        ...gameSession,
+        sessionId: uuid.v4(),
+        questionsAnswered: 0,
+        responseTimes: []
+      },
+      hasStarted: true
```
kevinrobinson: Would it make sense to have the session be created here, and then we don't need a separate state value for `hasStarted` since we can just check "is there a gameSession"?

ui/src/message_popup/experience_page.jsx
```diff
@@ -168,25 +157,27 @@ export default React.createClass({
...
-        itemsToShow={this.props.query}
-        helpType={this.state.helpType}
-        />);
+      <VelocityTransitionGroup enter={{animation: "callout.pulse", duration: 500}} leave={{animation: "slideUp"}} runOnMount={true}>
+        <div>
+          <InstructionsCard 
+           itemsToShow={this.props.query}
+           />
+          <ScaffoldingCard
+            email={gameSession.email}
```
kevinrobinson: Let's have the `ScaffoldingCard can own the email state, and not handing that information back to the experience page until it's wrapped up into a `gameSession`.

So this can become `initialEmail` to indicate it's like a "default value" but `ScaffoldingCard` tracks its own state.

ui/src/message_popup/experience_page.jsx
```diff
@@ -168,25 +157,27 @@ export default React.createClass({
...
+            scaffolding={scaffolding}
+            itemsToShow={this.props.query}
+            saveScaffoldingAndSession={this.onSaveScaffoldingAndSession}
```
kevinrobinson: Let's call something that describes what's happening conceptually, more than the pieces involved.  How about `onSessionStarted` or `onSessionConfigured`?  It's generally helpful to use the `on` prefix to indicate that something is a callback function, and this clarifies that this is only getting called at the end (not when the scaffolding values change).

ui/src/message_popup/popup_question.jsx
```diff
@@ -29,9 +29,11 @@ export default React.createClass({
...
 
   propTypes: {
-    shouldShowStudentCard: React.PropTypes.bool.isRequired,
-    shouldShowSummary: React.PropTypes.bool.isRequired,
-    helpType: React.PropTypes.string.isRequired,
+    scaffolding: React.PropTypes.shape({
+      helpType: React.PropTypes.string.isRequired,
+      shouldShowStudentCard: React.PropTypes.bool.isRequired,
+      shouldShowSummary: React.PropTypes.bool.isRequired
+    }).isRequired,
```
kevinrobinson: 👍

ui/src/message_popup/scaffolding_card.jsx
```diff
@@ -1,61 +1,149 @@
...
+    email: React.PropTypes.string.isRequired,
+    itemsToShow: React.PropTypes.object.isRequired,
+    saveScaffoldingAndSession: React.PropTypes.func.isRequired
   },
   
+  getInitialState(){
+    const isSolutionMode = _.has(this.props.itemsToShow, 'solution');
+    return ({
+      email: this.props.email,
+      questions: allQuestions,
```
kevinrobinson: We can remove `questions` from state altogether, since `getQuestions` uses the information in `competencyGroupValue` and `isSolutionMode` to compute them.

ui/src/message_popup/scaffolding_card.jsx
```diff
@@ -1,61 +1,149 @@
...
+      sessionLength: 10,
+      helpType: this.props.scaffolding.helpType,
+      shouldShowStudentCard: this.props.scaffolding.shouldShowStudentCard,
+      shouldShowSummary: this.props.scaffolding.shouldShowSummary,
```
kevinrobinson: It might be clearer to keep `scaffolding` packaged up as an object here.

ui/src/message_popup/scaffolding_card.jsx
```diff
@@ -1,61 +1,149 @@
...
+    if(competencyGroup !== ""){
+      competencyGroupValue = competencyGroup;
+    }
+    const questions = (isSolutionMode || competencyGroupValue === ALL_COMPETENCY_GROUPS)
+      ? _.shuffle(this.withStudents(allQuestions))
+      : this.questionsForCompetencies(competencyGroupValue);
+    return questions;
+  },
+  
+  withStudents(questions) {
```
kevinrobinson: You can import this from `transformations.jsx` and delete this copy here.

ui/src/message_popup/scaffolding_card.jsx
```diff
@@ -1,61 +1,149 @@
...
+    return questions.map((question) => {
+      const student = _.find(allStudents, {id: question.studentId });
+      return _.extend({student}, question);
+    });
+  },
+  
+  questionsForCompetencies(competencyGroup) {
```
kevinrobinson: ditto!  You can import this from `transformations.jsx` and delete this copy here.

ui/src/message_popup/scaffolding_card.jsx
```diff
@@ -1,61 +1,149 @@
...
+      if (learningObjective.competencyGroup !== competencyGroup) return null;
+      return {
+        ...question,
+        competencyGroup: learningObjective.competencyGroup
+      };
+    }));
+    return _.shuffle(this.withStudents(withCompetencyGroups));
+  },
+
+  onSave(){
```
kevinrobinson: This looks great!

ui/src/message_popup/scaffolding_card.jsx
```diff
@@ -1,61 +1,149 @@
...
+    const scaffolding = {
+      helpType: this.state.helpType,
+      shouldShowStudentCard: this.state.shouldShowStudentCard,
+      shouldShowSummary: this.state.shouldShowSummary
+    };
+    const gameSession = {
+      email: this.state.email,
+      sessionLength: this.state.sessionLength,
+      questions: this.getQuestions(this.state.competencyGroupValue)
```
kevinrobinson: I don't think any other components need to know about `sessionLength` - ultimately it's only used to produce a list of questions for the gameSession.  If we do something like `questions.slice(0, sessionLength)` then we can pass back just the `email` and the list of questions to the parent.

ui/src/message_popup/scaffolding_card.jsx
```diff
@@ -1,61 +1,149 @@
...
+  },
   
+  onCompetencyGroupChanged(event, competencyGroupValue){
+    const questions = this.getQuestions(competencyGroupValue);
+    const newLength = questions.length;
+    var sessionLength = this.state.sessionLength;
+    if(sessionLength > newLength){
+      sessionLength = newLength;
+    }
+    this.setState({questions, sessionLength, competencyGroupValue});
```
kevinrobinson: I don't think we need to save the `questions` as state - they can be computed from the other state values when we need it (like in `onSave`).

ui/src/message_popup/scaffolding_card.jsx
```diff
@@ -1,61 +1,149 @@
...
   },
+
+  onTextChanged({target:{value}}:TextChangeEvent) {
+    this.setState({email: value});
+  },
   
   render(){
-    const {competencyGroupValue, competencyGroups, onCompetencyGroupChanged, sessionLength, questions, onSliderChange, shouldShowStudentCard, onStudentCardsToggled, shouldShowSummary, onSummaryToggled, helpType, onHelpToggled, itemsToShow} = this.props;
+    const {itemsToShow} = this.props;
+    const {competencyGroupValue, sessionLength, questions, shouldShowStudentCard, shouldShowSummary, helpType} = this.state;
```
kevinrobinson: If we remove `questions` you can compute them by calling `getQuestions` here.

ui/src/message_popup/scaffolding_card.jsx
```diff
@@ -1,61 +1,149 @@
...
-    var showStudentCardsToggle = _.has(itemsToShow, 'all') || _.has(itemsToShow, 'cards') || _.has(itemsToShow, 'basic');
-    var showSummaryToggle = _.has(itemsToShow, 'all') || _.has(itemsToShow, 'summary');
-    var showHelpToggle = _.has(itemsToShow, 'all') || _.has(itemsToShow, 'feedback') || _.has(itemsToShow, 'basic');
-    var showOriginalHelp = _.has(itemsToShow, 'all') || _.has(itemsToShow, 'originalHelp');
+    const showLearningObjectives = true;
+    const showSlider = true;
+    const showStudentCardsToggle = _.has(itemsToShow, 'all') || _.has(itemsToShow, 'cards') || _.has(itemsToShow, 'basic');
+    const showSummaryToggle = _.has(itemsToShow, 'all') || _.has(itemsToShow, 'summary');
+    const showHelpToggle = _.has(itemsToShow, 'all') || _.has(itemsToShow, 'feedback') || _.has(itemsToShow, 'basic');
+    const showOriginalHelp = _.has(itemsToShow, 'all') || _.has(itemsToShow, 'originalHelp');
```
kevinrobinson: This looks like something we could simplify, but that's something for another day :)

david-rosales:
@kevinrobinson I made more revisions, can you review when you get a chance?

ui/src/message_popup/experience_page.jsx
```diff
@@ -46,42 +42,19 @@ export default React.createClass({
...
-      email: this.context.auth.userProfile.email,
-      hasStarted: false,
-      questionsAnswered: 0,
-      sessionLength: 10,
+      scaffolding: {
+        helpType: isSolutionMode ? 'none' : 'feedback',
+        shouldShowStudentCard: isSolutionMode ? _.has(this.props.query, 'cards') : true,
+        shouldShowSummary: !isSolutionMode,
+      },
       toastRevision: false,
```
kevinrobinson: It'd be good to add in `gameSession: null`, to just to indicate that's part of the state but isn't set yet when the component first mounts.

ui/src/message_popup/experience_page.jsx
```diff
@@ -91,56 +64,69 @@ export default React.createClass({
...
+    this.setState({ gameSession });
   },
 
   onDonePressed() {
     Routes.navigate(Routes.Home);
   },
   
   resetExperience(){
-    this.setState(this.getInitialState());
+    this.replaceState(this.getInitialState());
```
kevinrobinson: I'm assuming this is to remove `gameSession`?  It's good to explicitly set that to null anyway, and then this will work as the regular `setState` as well.

kevinrobinson:
@david-rosales Approved!  There ended up being a bit involved here, so thanks for all your work to help simplify. 👍  It will help make it more direct to change things like "what scaffolding should be setup" and "what are different sets of questions in a gameSession."

PR #70, opened 2016-07-06
Improve scrolling with animated transition in scoring UI

kevinrobinson:
This is a small improvement to set the scroll position to the top of the window when navigating between the scoring screens.  It runs after the Velocity transition animation finishes.

This also has some unrelated eslint cleanup :)


kevinrobinson:
@david-rosales Following your great idea! :)

kevinrobinson:
@david-rosales Can you review this when you have a chance?

david-rosales:
@kevinrobinson Approved!

PR #71, opened 2016-07-11
Fix navigation from when done a question in scoring UI

kevinrobinson:
This was a bug from before, now the "All done" button that's shown when you finish scoring all the responses for a question works as expected.


kevinrobinson:
@david-rosales Can you review this small fix?

david-rosales:

PR #72, opened 2016-07-11
Fix bug threading through default sessionLength

kevinrobinson:
@david-rosales I wanted to change this for the playtest and the `sessionLength` wasn't being threaded through.  I think this is something we'll want to simplify as part of the refactoring in https://github.com/mit-teaching-systems-lab/threeflows/pull/67#issuecomment-230902000 as well.


kevinrobinson:
@david-rosales Can you review this change as well?

kevinrobinson:
@david-rosales Selfie-ing this to get it ready for tomorrow.

PR #68, opened 2016-07-06
Add in indicators to scoring UI

kevinrobinson:
screen shot 2016-07-06 at 2 27 06 pm

screen shot 2016-07-06 at 2 27 17 pm

I also ran a corresponding migration manually.

PR #74, opened 2016-07-11
Fix bug with questions mapped to indicators

kevinrobinson:
This fixes a bug from https://github.com/mit-teaching-systems-lab/threeflows/pull/68.


kevinrobinson:
@david-rosales selfie-ing to fix this bug.

PR #77, opened 2016-07-11
Fix bug with threading helpType through to instructions page

kevinrobinson:
The `helpType` read from the URL in `experience_page` isn't passed into the `instructions_card`, so it essentially has no effect on the experience.  

@david-rosales this is also likely related to the refactoring like https://github.com/mit-teaching-systems-lab/threeflows/pull/67#issuecomment-230902000


kevinrobinson:
Going to selfie this to get it ready for tomorrow  cc @david-rosales.

PR #78, opened 2016-07-12
Do scoring on separate domain with isolated problems responses

kevinrobinson:
I stood up a separate domain for scoring, so that the items that folks are scoring during the playtest are controlled past responses.  Responses from folks playing are kept separate and not visible to other players.


kevinrobinson:
selfie cc @david-rosales

PR #79, opened 2016-07-12
Add Rollbar to report JS errors

kevinrobinson:
So we have visibility into browser-side JS errors.  The environment var here is set based on the deployment so that we can collect all the things for now.  We can adjust this or turn off local errors if we find this bothersome.


kevinrobinson:
@david-rosales I also invited you to the Rollbar org, so we'll get alerted for JS errors that occur in production (and specifically so I'll get alerted for any  during the playtest today).

PR #80, opened 2016-07-12
Add eslint and Flow to build process

kevinrobinson:
This adds in eslint and Flow to the `npm test` command, which will also cause it to be run as part of the Travis build process.  The checks are run sequentially and a failure in either will break the build.


kevinrobinson:
@david-rosales This will start running these checks in the Travis build, so if you can run these locally on your sweet new computer, we should be ready for this!  Can you take a look and confirm?

david-rosales:
@kevinrobinson Works great on my computer, approved!

kevinrobinson:
@david-rosales sweet, thanks! 👍

PR #83, opened 2016-07-12
Adding in a Text Interface for Message PopUp

david-rosales:
I implemented the MobilePrototype of Message PopUp. It looks like a texting interface and allows the user to respond to situations through 'text messaging'. Most of the features from the non Mobile Prototype version are available except for the summary between questions. I felt that the initial point of the summary was to have a pause for the user and show key information. However, the pause already occurred in the Mobile Prototype with users having to press the 'Next Question' button before moving on. Also, while the 'Next Question' button is displayed, the timer stops but is also still displayed. The chatlog also remains so all the key information was present and an extra box with a summary looked very redundant. There are a lot of props being passed into the `MobileInterface` and I decided to leave it like that for now because I want to handle the refactoring in a separate pull request because it will involve similar changes to a refactor in the `ExperiencePage`.

Below is a simple screenshot of one of the changes from the previous prototype. The change is that student info is no longer shown as a dialog when pressing on the face icon for the student because that feature might not seem intuitive. Instead, it shows up as a separate message bubble.
screen shot 2016-07-12 at 1 11 59 pm

There is now also a feedback and revision option. 
screen shot 2016-07-12 at 1 12 49 pm

Once the initial response is sent, the buttons on the bottom change.
screen shot 2016-07-12 at 1 43 09 pm

Also added in the show example button.
screen shot 2016-07-12 at 1 21 57 pm

Examples are sent as messages until all the examples have been gone through.
screen shot 2016-07-12 at 1 22 13 pm

Once the examples have all been gone through, the 'Show Example' button becomes disabled.
screen shot 2016-07-12 at 1 41 46 pm

Instead of a toast, an info message is now sent when the user passes the 90 second mark.
screen shot 2016-07-12 at 1 44 30 pm

Lastly, I decided to change the color of the 'Next Question' button to make it more obvious that the button needs to be pressed to continue. Before, the button would blend into the previous 'Send' button or 'Save Response' button.
screen shot 2016-07-12 at 1 13 16 pm

A lot of the props will be simplified in the future along with some of the functions within the components.


david-rosales:
@kevinrobinson There's a lot of additions, can you review this when you get a chance?

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,383 @@
...
+          }
+        </div>
+      </div>
+    );
+  }
+});
+
+
+const styles = {
+  appBar: {
```
kevinrobinson: Do these all need to be absolutely positioned?  We should try to do some flexbox-ifying :)  I can pair with you on that too if you like.

ui/src/message_popup/experience_page.jsx
```diff
@@ -26,6 +26,8 @@ import * as Api from '../helpers/api.js';
 import FinalSummaryCard from './final_summary_card.jsx';
 import InstructionsCard from './instructions_card.jsx';
 
+import TextingInterface from './mobile_prototype/texting_interface.jsx';
```
kevinrobinson: It looks like this isn't used anymore?

ui/src/message_popup/experience_page.jsx
```diff
@@ -212,11 +214,47 @@ export default React.createClass({
...
+      questionsAnswered,
+      shouldShowStudentCard,
+      shouldShowSummary,
+      helpType,
+      sessionLength
+    } = this.state;
+    const question = questions[questionsAnswered];
+    return ( 
+      <MobileInterface
+        hasStarted = {this.state.hasStarted}
```
kevinrobinson: As you were saying in person, it'd be good to land the refactoring in https://github.com/mit-teaching-systems-lab/threeflows/pull/67 first so that this can become a simpler interface too.

ui/src/message_popup/experience_page.jsx
```diff
@@ -212,11 +214,47 @@ export default React.createClass({
...
+        helpType={helpType}
+        limitMs={this.state.limitMs}
+        onLog={this.onLog}
+        resetExperience={this.resetExperience}
+        isLastQuestion={questionsAnswered+1===sessionLength ? true : false}
+        responseTimes={this.state.responseTimes}
         />
     );
+    
+    /*return (
```
kevinrobinson: This is leftover, right?

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,383 @@
...
+  },
+  
+  onStart(email, sessionLength, questions, shouldShowStudentCard,shouldShowSummary, helpType){
+    this.props.onStartPressed(email, sessionLength, questions, shouldShowStudentCard,shouldShowSummary, helpType);
+  },
+  
+  addMessages(messageList){
+    var messages = _.clone(this.state.messages);
+    var initialResponse = this.state.initialResponse;
+    var revisedResponse = this.state.revisedResponse;
```
kevinrobinson: Switch these `var`s over to `const`s where you can do indicate that they don't get mutated at all.  You can also use destructuring like this to simplify a bit:

```
const {initialResponse, revisedResponse} = this.state;
```

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,383 @@
...
+    for(var messageIndex = 0; messageIndex < messageList.length; messageIndex++){
+      this.messageId+=1;
+      const message = _.extend({key: this.messageId}, messageList[messageIndex]);
+      if(message.text !== ''){
+        messages.push(message);
+      }
+      if(message.type === 'user'){
```
kevinrobinson: This looks like it takes a list of messages and then extracts whether there was an `initialResponse` or `revisedResponse` in there.  Could this be done at the call site?

In order words, in the method handling when the user types in the revised response message, would it work to have that `setState` to add a message and then also setState for the `revisedResponse`?

Alternately we could use only `messages` with the different types you have and write a method to pull out the last `revisedResponse` and remove that from state altogether.  Essentially, keep all that state in messages.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,383 @@
...
+    };
+    this.props.onLog(type, response);
+  },
+
+  resetExperience(){
+    this.setState(this.getInitialState());
+    this.props.resetExperience();
+  },
+
+  getInitialMessages(nextProps){
```
kevinrobinson: I'm not quite following what this method is doing, is it checking the props and then figuring out what initial messages to show based on the question?

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,383 @@
...
+  },
+
+  updateTimer() {
+    if(!this.isReadyToMoveOn()){
+      this.setState({ elapsedMs: this.state.elapsedMs + ONE_SECOND });
+    }
+  },
+  
+  componentWillMount(){
+    this.messageId = parseInt(this.props.question.id)*10 + 1;
```
kevinrobinson: I'm not quite sure what this is about - do messages need an id in order to use it as a key when rendering?  I was thinking that the message index might be sufficient for that purpose.  Or, if you want to make a unique message id, you can used the `uuid` package.  I'm not sure why it's related to the `question.id` here though.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,383 @@
...
+  },
+  
+  componentWillReceiveProps(nextProps){
```
kevinrobinson: In general, React components shouldn't use `componentWillReceiveProps` and `componentDidUpdate`.  These are primarily for interop with other libraries or dropping down to do raw DOM manipulations.  You shouldn't have to do that in most normal UI development.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,383 @@
...
+    var messages = this.getInitialMessages(nextProps);
+    const goodExamples = _.map(nextProps.question.examples, (example) => {
+      return {type: 'Good', text: example};
+    });
+    const badExamples = _.map(nextProps.question.nonExamples, (example) => {
+      return {type: 'Bad', text: example};
+    });
+    const allExamples = _.shuffle(_.clone(goodExamples).concat(badExamples));
```
kevinrobinson: Why are the examples getting re-shuffled here?

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,383 @@
...
+  componentWillReceiveProps(nextProps){
+    var messages = this.getInitialMessages(nextProps);
+    const goodExamples = _.map(nextProps.question.examples, (example) => {
+      return {type: 'Good', text: example};
+    });
+    const badExamples = _.map(nextProps.question.nonExamples, (example) => {
+      return {type: 'Bad', text: example};
+    });
+    const allExamples = _.shuffle(_.clone(goodExamples).concat(badExamples)); 
+    //this.addMessages(messages);
```
kevinrobinson: leftover comment

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,383 @@
...
+    const elapsedMs = 0;
+    this.setState({ messages, allExamples, elapsedMs: 0 });
+  },
+  
+  componentDidUpdate(prevProps, prevState){
+    if(prevState.elapsedMs === 90000){
+      this.addMessages([{type: 'info', text: 'Remember, these are supposed to be quick responses.'}]);
```
kevinrobinson: This seems more natural to put in `updateTimer`, rather than after every state update.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,383 @@
...
+    }
+    if(prevState.initialResponse === undefined && this.state.initialResponse !== undefined){
+      this.logResponse();
+    }else if(prevState.revisedResponse === undefined && this.state.revisedResponse === ''){
+      this.logRevisionDeclined();
+    }else if(prevState.revisedResponse === undefined && this.state.revisedResponse !== undefined){
+      this.logRevision(this.state.revisedResponse);
+    }
```
kevinrobinson: Let's see if we can do this at the call site, where we handle the action that the user is taking, instead of having to compute what has changed here.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,383 @@
...
+    return (
+      <div>
+        <AppBar
+          style={styles.appBar}
+          title="Message PopUp Mobile"
+          iconElementLeft={
+            <IconButton
+              onTouchTap={this.resetExperience}
+            >
+              {!hasStarted &&
```
kevinrobinson: This seems like it could be more readable as one line since these are so short and wouldn't line-wrap, up to you.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,383 @@
...
+              {hasStarted && questionsAnswered >= sessionLength &&
+                <RefreshIcon 
+                  color={fullWhite}
+                />
+              }
+            </IconButton>
+          }
+          />
+        <div>
+        {/*Header*/}
```
kevinrobinson: leftover comment

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,383 @@
...
+          
+        </div>
+        <div>
+        {/*Scollable Body*/}
```
kevinrobinson: leftover comment

ui/src/message_popup/mobile_prototype/texting_interface.jsx
```diff
@@ -0,0 +1,154 @@
+import React from 'react';
+
+import Divider from 'material-ui/Divider';
+import TextField from 'material-ui/TextField';
+import RaisedButton from 'material-ui/RaisedButton';
+
+import {StudentMessage, UserMessage, InfoMessage} from './message_components.jsx';
+
+export const TextBody = React.createClass({
+  propTypes: {
```
kevinrobinson: Please add in a `displayName` property for all React components so they look nice in debugging messages.

ui/src/message_popup/mobile_prototype/texting_interface.jsx
```diff
@@ -0,0 +1,154 @@
...
+    this.props.addResponse(this.state.text);
+    this.setState({initialMessageSent: true});
+  },
+  
+  onRevisionSendMessage(){
+    this.onSendMessage();
+  },
+  
+  onPassSendMessage(){
+    this.props.addResponse('');
```
kevinrobinson: This seems like it's a different user action and important to capture - passing on revising vs. adding a blank response.

ui/src/message_popup/mobile_prototype/texting_interface.jsx
```diff
@@ -0,0 +1,154 @@
...
+  },
+  
+  onNextQuestion(){
+    this.setState(this.getInitialState());
```
kevinrobinson: In general this isn't a great pattern in React - it's better to update state and then let new components be created and destroyed based on what `render` returns rather than managing the state changes like this manually.  That's one of the main benefits, that you often don't have to think of resetting state.

Here, if we remove this what would happen?

kevinrobinson:
@david-rosales Nice work on taking an initial product concept, sketching it out and then prototyping parts of it!

Here's my suggestions for next steps, echoing some of what we talked about in person:
1. there's too much here to try to get this merged for the playtest tonight, which is fine
2. let's finish the refactoring work in https://github.com/mit-teaching-systems-lab/threeflows/pull/67 first
3. let's try to clarify the central texting metaphor in some design sketches next
4. then we can do an internal playtest on those sketches

In terms of the code here, I think it'd be good to sync on `componentDidUpdate` and `componentDidReceiveProps` and the flow of state through the components.  Let's do that in person whenever you think is a good time.

kevinrobinson:
@david-rosales Thanks for sharing this and talking it out in person, your screenshots look like they're ready for some playtesting!  👍 

You can put them in a Slides presentation and then show people in person one-on-one and get feedback.  Then you can take the input from a few people in person and synthesize it to make your plan for what the next steps should be.

david-rosales:
I thought that I'd document a few of the changes I've done to the Mobile Prototype since the initial build. One of them is that student information is now displayed by touching on the student's icon or through an information icon.

Touching on an information icon will bring up this Dialog:
screen shot 2016-07-28 at 11 43 44 am
The reasoning behind using the information icon as well is because not all questions will have students speaking. Some will solely be situational while there still is a student involved. In those cases, accessing the student's information is still important.

From the Involved Students Dialog, you can tap on a student chip or from the texting interface, you can tap on a student's icon to display their information like below:
screen shot 2016-07-28 at 11 44 19 am

Another addition is that there is now support for multiple students. I originally added this feature here but as it came up in a separate pull request I decided to redo it on its own branch so that it could be shared between in progress pull requests. The difference here being that there is `encodedText` which works as follows:
- If the text is informational, enter the following tag before it: `[:i]`.
- If the text is from the user, enter the following tag before it: `[:u]`.
- If the text is from a student, enter the following tag before it: `[:{studentId}s]`. For example, if Ada was the student, Ada's ID is 8 and so the tag would be `[:8s]`.
- The text after the tag will be in the type until the next tag.

With multiple students also comes the feature that students now speak what they "say" in the question. The `encodedText` is a property of a question just as `text` or `id` is. Furthermore, the `studentId` Number property has been changed to a `studentIds` Array property to support multiple students.

Below are two examples of `encodedText` in action:
<div style="display: flex; justify-content: space-between">
screen shot 2016-07-28 at 11 48 40 am
screen shot 2016-07-28 at 11 49 19 am
</div>
The `encodedText` for the first image is:
```
encodedText: '[:i]You introduced a game related to photosynthesis, and have pairs of two working on a game. Maia is paired with Kevin.[:3s]How are we getting graded on this?[:u]There will be no grade for either completion or participation.[:3s]Why are we doing this then?'
```
The `encodedText` for the second image is:
```
encodedText: '[:i]At the end of the photosynthesis unit, you introduce a list of topics that students choose for a project they want to do. Maia raises her hand.[:3s]I don\'t want to do any of these. Can I just come up with my own thing?[:i]Even before you have a chance to respond to her, another student raises her hand. [:8s]I want to do my own topic too!'
```

Another thing that was added because of the multiple students is names near the top of each message bubble. This way, if there are multiple students speaking, like in the second image above, even though the students have the same icon, they are still differentiable.
screen shot 2016-07-28 at 11 50 18 am


Furthermore, since there are multiple students, you can also see multiple chips in the Involved Students Dialog.
<div style="display: flex; justify-content:space-between; flex-flow: column || nowrap;">
screen shot 2016-07-28 at 11 50 55 am
screen shot 2016-07-28 at 12 28 49 pm
</div>

Another issue I ran into was how to communicate to users that tapping on icons did things. One suggestion was to state it somewhere on the instructions page.
screen shot 2016-07-28 at 11 54 41 am
 The only issue with this approach is that most people regardless of how much it is stressed to do so, don't read instructions. So in order to make it clearer that tapping on icons was a feature, I added in a Final Reminders Dialog that appears when users click start.
screen shot 2016-07-28 at 11 55 19 am
This should be enough to communicate the necessary information to users. However, one other suggestion I received that I experimented with was making the icons look like they were clickable by giving them a subtle animation. I tried this but ran into the issue of trying to decide which icons to animate. Also, if I was animating every icon that was clickable, all the icons on the left would be animated and in unison. The movement in unison was extremely distracting and I think takes away from the overall experience so I did not pursue it. However, if we find that the above two attempts at attracting user attention are insufficient, we can try again with different animations.

On another note, animations have been added to the messages so that they appear to pop in when "sent". 

Lastly, I updated the feedback view to be solely information text (compared to the change between information and student text that was originally implemented). The feedback/revision message now looks as follows:
screen shot 2016-07-28 at 11 53 58 am

One other thing we need to do which can be in a separate pull request is, for the new questions, assign students that are in the virtual classroom.

This is an update/overview of the changes since the first review. I might have missed a few small things and will add them if I recall/come across them.

ui/src/message_popup/experience_page.jsx
```diff
@@ -127,9 +136,16 @@ export default React.createClass({
               primaryText="Restart game" />
           }
         />
+        <Dialog
```
kevinrobinson: We talked about removing this, but also as it is written now this would affect the simple "non-mobile prototype" interface too.  When making larger changes like this it's best if we can isolate the changes to only the new UI code, like within `renderMobilePrototype` or code it uses.

ui/src/message_popup/mobile_prototype/message_components.jsx
```diff
@@ -0,0 +1,178 @@
...
+import ReactDOM from 'react-dom';
+import _ from 'lodash';
+
+import Paper from 'material-ui/Paper';
+import IconButton from 'material-ui/IconButton';
+
+import FaceIcon from 'material-ui/svg-icons/action/face';
+import InfoOutlineIcon from 'material-ui/svg-icons/action/info-outline';
+
+const Message = React.createClass({
```
kevinrobinson: This is a great use of the `Message` component as a small piece that other components use. 👍 

If you choose to have multiple components in a file like this, please add a quick comment at the top of the file about what this file contains and what's going on.

ui/src/message_popup/mobile_prototype/message_components.jsx
```diff
@@ -0,0 +1,178 @@
...
+    type: React.PropTypes.string.isRequired,
+    text: React.PropTypes.string.isRequired,
+    messageStyle: React.PropTypes.object.isRequired,
+    messageTextStyle: React.PropTypes.object.isRequired,
+    label: React.PropTypes.string,
+    onOpenDialog: React.PropTypes.func
+  },
+  
+  componentDidMount(){ 
+    ReactDOM.findDOMNode(this).scrollIntoView();
```
kevinrobinson: Will this work as expected if multiple `Messages` are on the page?

ui/src/message_popup/mobile_prototype/message_components.jsx
```diff
@@ -0,0 +1,178 @@
...
+  componentDidMount(){ 
+    ReactDOM.findDOMNode(this).scrollIntoView(); 
+  },
+  
+  render(){
+    const {type, messageStyle, messageTextStyle} = this.props;
+    
+    return (
+      <div style={messageStyle}>
+        { type === 'student' &&
```
kevinrobinson: It seems like the core idea here is that `Message` can have a different icon, or maybe no icon.  Instead of putting the details of the `StudentMessage` and `InfoMessage` inside here, could you update `Message` to take an `icon` and then pass in those values from the particular message components?

ui/src/message_popup/mobile_prototype/message_components.jsx
```diff
@@ -0,0 +1,178 @@
...
+  message: {
+    display: 'flex',
+    padding: 5
+  },
+  messageIconButton: {
+    margin: 0,
+    padding: 0,
+    flexGrow: 0,
+    flexShrink: 0,
+    flexBasis:30
```
kevinrobinson: Does it make sense for this and the other icons to have a fixed width and height?

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,375 @@
...
+const ONE_SECOND = 1000;
+
+export default React.createClass({
+  displayName: 'MobileInterface',
+
+  mixins: [SetIntervalMixin],
+  
+  getInitialState() {
+    var allExamples = [];
+    var messages = [];
```
kevinrobinson: Can `allExamples` and `messages` be constant?

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,375 @@
...
+    const goodExamples = _.map(this.props.question.examples, (example) => { return {type: 'Good', text: example}; });
+    const badExamples = _.map(this.props.question.nonExamples, (example) => { return {type: 'Bad', text: example}; });
+    allExamples = _.shuffle(_.clone(goodExamples).concat(badExamples)); 
+    messages = this.getInitialMessages(this.props.question, this.props.scaffolding.shouldShowStudentCard);
+    return ({
+      messages,
+      allExamples,
+      selectedStudent: this.props.question.students[0],
```
kevinrobinson: Is this just used for the dialog?  Should it be part of that, something like `setState({ dialog: { type: 'student', studentId: 5 })`?

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,375 @@
...
+      initialResponse: undefined,
+      revisedResponse: undefined,
+      elapsedMs: 0,
+      exampleIndex: 0,
+      dialog: 'none',
```
kevinrobinson: Would `null` make sense here?

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,375 @@
...
+      messageId: 1
```
kevinrobinson: This looks like an index into the `messages` array rather than a primary key that you'd use to look up a message, so i'd suggest `messageIndex` instead.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,375 @@
...
+    var messageId = this.state.messageId;
+    for(var messageIndex = 0; messageIndex < messageList.length; messageIndex++){
+      messageId+=1;
+      const message = _.extend({key: messageId}, messageList[messageIndex]);
+      messages.push(message);
+    }
+    this.setState({messages, messageId});
+  },
+  
+  addResponse(response){
```
kevinrobinson: This can make sense as a method, but if it's for adding `user` messages then the name should reflect that and you'll want to make sure you call it each time.  As is, some methods use this to add a user message but others create them directly, so I'd either update to have them all use this or remove it.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,375 @@
...
+  sendRevision(response){
+    if(this.props.question.examples.length > 0){
+      this.addMessages([
+        {type: 'user', text: response},
+        {type: 'info', text: "Here's an example of something you could say/do:\n\n" + _.shuffle(this.props.question.examples)[0] + "\n\nWould you like to revise your response?"}, 
+      ]);
+    }else{
+      this.addMessages([
+        {type: 'user', text: response},
+        {type: 'info', text: "There aren't any examples for this situation yet but your response will be recorded for feedback."},
```
kevinrobinson: I'd cut this line altogether, there may not actually be feedback for many of the testing sessions.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,375 @@
...
+    }else{
+      this.addMessages([
+        {type: 'user', text: response},
+        {type: 'info', text: "There aren't any examples for this situation yet but your response will be recorded for feedback."}, 
+      ]);
+      this.setPassedResponse();
+    }
+  },
+  
+  showExample(){
```
kevinrobinson: As a convention, please update this method and others in this component to indicate they're a response to an event.  For this it might be `onShowExampleClicked`.  This can help clarify which methods are event handlers and which ones are side-effect free and just calculate something from state (eg., `isReadyToMoveOn`).

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,375 @@
...
+      if(increment){
+        this.setState({exampleIndex: this.state.exampleIndex + 1});
+      }
+      return this.state.allExamples[currentIndex];
+    }else{
+      return null;
+    }
+  },
+
+  setDialog(value){
```
kevinrobinson: I'd remove this, I'm not sure it's adding much to say `setDialog(value)` over `setState({dialog: value})`.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,375 @@
...
+
+  onOpenInfoDialog(){
+    this.setDialog('info');
+  },
+
+  onOpenStudentDialog(student){
+    return (function() {
+      this.setState({selectedStudent: student});
+      this.setDialog('student');
+    }.bind(this));
```
kevinrobinson: Do you need this `bind` here?

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,375 @@
...
+      shouldShowStudentCard,
+      helpType,
+      elapsedMs,
+      initialResponseText: initialResponse,
+      finalResponseText
+    };
+    this.props.onLog(type, response);
+  },
+
+  getInitialMessages(questionObject, shouldShowStudentCard){
```
kevinrobinson: Does this use `shouldShowStudentCard`?  Looks like the linter settings don't pick this up and they should, I'll update that separately.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,375 @@
...
+    var messages = [];
+    if(_.has(questionObject, 'encodedText')){
+      var arrayOfMessages = questionObject.encodedText.split('[:');
```
kevinrobinson: Encoding this structure in the questions directly will let you remove this parsing code altogether :)

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,375 @@
...
+        </div>
+        <div>
+          <div> 
+            <div style={styles.textFooter}>
+              <TextFooter 
+                onQuestionDone={this.onNextQuestion}
+                setInitialResponse={this.setInitialResponse}
+                setRevisedResponse={this.setRevisedResponse}
+                setPassedResponse={this.setPassedResponse}
+                log={{logResponse: this.logResponse, logRevision: this.logRevision, logRevisionDeclined: this.logRevisionDeclined}}
```
kevinrobinson: Nice wrapping this up into a logical object!

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -0,0 +1,375 @@
...
+                isReadyToMoveOn={this.isReadyToMoveOn}
+                helpType={helpType}
+                elapsedMs={this.state.elapsedMs}
+                showExample={this.showExample}
+                nextExample={this.nextExample}
+                nextButtonLabel={this.props.isLastQuestion ? 'Finish' : 'Next Question'}
+                />
+            </div>
+            <div>
+              {selectedStudent !== undefined &&
```
kevinrobinson: This looks good, but I'd recommend splitting out the rendering of the dialogs to their own method.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
kevinrobinson: Ping on this question.

ui/src/message_popup/mobile_prototype/texting_interface.jsx
```diff
@@ -0,0 +1,184 @@
...
+import VelocityComponent from 'velocity-react/velocity-component';
+import 'velocity-animate/velocity.ui';
+
+import Divider from 'material-ui/Divider';
+import TextField from 'material-ui/TextField';
+import RaisedButton from 'material-ui/RaisedButton';
+
+import {StudentMessage, UserMessage, InfoMessage} from './message_components.jsx';
+
+export const TextBody = React.createClass({
```
kevinrobinson: Would it work to split these two components into their own files?  If not please add a quick comment at the top of the file explaining what's in here.

ui/src/message_popup/mobile_prototype/texting_interface.jsx
```diff
@@ -0,0 +1,184 @@
...
+  displayName: 'TextBody',
+
+  propTypes: {
+    question: React.PropTypes.object.isRequired,
+    messages: React.PropTypes.array.isRequired,
+    onOpenStudentDialog: React.PropTypes.func.isRequired,
+    onOpenInfoDialog: React.PropTypes.func.isRequired
+  },
```
kevinrobinson: Awesome API to the component, super clear!

ui/src/message_popup/mobile_prototype/texting_interface.jsx
```diff
@@ -0,0 +1,184 @@
...
+  
+  render(){
```
kevinrobinson: Nice!  This is great factoring to pull out the components of the messages here, it makes this clear and direct.

ui/src/message_popup/mobile_prototype/texting_interface.jsx
```diff
@@ -0,0 +1,184 @@
...
+    const {messages} = this.props;
+    return (
+      <div>
+        {_.map(messages, (message) => {
```
kevinrobinson: You can use `message.map` directly, Arrays have map built in.

ui/src/message_popup/mobile_prototype/texting_interface.jsx
```diff
@@ -0,0 +1,184 @@
...
+        <Divider />
+        <div style={styles.responseTextBox}>
+          <TextField 
+            id="Response"
+            multiLine={true}
+            underlineShow={false}
+            rows={2}
+            rowsMax={2}
+            onChange={this.onTextChange}
+            disabled={this.props.isReadyToMoveOn()}
```
kevinrobinson: Instead of passing down the function `isReadyToMoveOn`, pass the value directly.

ui/src/message_popup/mobile_prototype/texting_interface.jsx
```diff
@@ -0,0 +1,184 @@
...
+    marginTop: 20,
+    marginLeft: 15,
+    marginRight: 15,
+  },
+  responseButton: {
+    marginLeft: 5,
+    marginRight: 5
+  },
+  responseTimer: {
+    position: 'fixed',
```
kevinrobinson: Can you use flexbox here?  I don't think we want fixed position for anything if we can avoid it.

kevinrobinson:
@david-rosales Awesome work!  I left some comments, let me know if you want to chat about any of them or if it's helpful for me to do a final review on this pass.

kevinrobinson:
@david-rosales Also, here are the raw notes from our earlier in-person conversation, which we broke up into three pull requests:

```
prompt about tapping on students: secondary so remove
end button -> respond to Maia
...
encodedText - encode structure directly
...
animation to side not grow in, could use stagger
```

ui/src/message_popup/mobile_prototype/message_components.jsx
```diff
@@ -147,7 +151,7 @@ const styles = {
     padding: 0,
     flexGrow: 0,
     flexShrink: 0,
-    flexBasis:30
+    flexBasis: 30
   },
   messageIcon: {
     width: 30,
```
david-rosales: I tried changing the fixed width and height by relying on setting just the flexBasis and using margins to position it but the IconButton wave effect looks really awkward and was off-centered.

david-rosales:
@kevinrobinson I went through the changes you suggested, removed the Final Reminders dialog, and changed the `SEND` button to be more action oriented. For now, there are a few questions that, although they should, do not have studentIds associated with them. With no studentIds, the `SEND` button now defaults to `RESPOND TO CLASS` instead of `RESPOND TO [StudentNameHere]`. Can you review the new changes one last time?

ui/src/message_popup/mobile_prototype/message_components.jsx
```diff
@@ -0,0 +1,182 @@
...
+  messageIconButton: {
+    margin: 0,
+    padding: 0,
+    flexGrow: 0,
+    flexShrink: 0,
+    flexBasis: 30
+  },
+  messageIcon: {
+    width: 30,
+    height: 30,
```
david-rosales: I tried changing the fixed width and height by relying on setting just the flexBasis and using margins to position it but the IconButton wave effect looks really awkward and was off-centered.

ui/src/message_popup/mobile_prototype/text_footer_component.jsx
```diff
@@ -0,0 +1,145 @@
...
+    this.props.setPassedResponse('');
+    this.props.log.logRevisionDeclined();
+  },
+  
+  onNextQuestion(){
+    this.props.onQuestionDone();
+  },
+  
+  render(){
+    const mainButtonLabel = this.props.isReadyToMoveOn ? this.props.nextButtonLabel : this.props.helpType !== 'feedback' ? 'Respond to ' + (this.props.mainStudent !== undefined ? this.props.mainStudent.name : 'Class') : this.state.initialMessageSent ? 'Revise Response' : 'Save Response';
```
kevinrobinson: This is a long line :)  Can you break this up a bit? Triply nested ternaries are usually harder to read than they're worth and there's a bug with the initial button text in practice mode at the moment that's hard to spot.

I'd suggest splitting this out into a method, and then checking for conditions that you can return early from.  Something like:

```
renderButtonText(): {
  if (this.props.isReadyToMoveOn) return this.props.nextButtonLabel;
  if (this.state.initialMessageSent && this.props.helpType === 'feedback') return 'Revise response';
  ...
}
```

kevinrobinson:
@david-rosales This looks great!  There's a small bug in the branching for different button texts, and then a merge conflict to resolve but then let's ship this.  Awesome work, this seems like an appropriate approval gif:

spiderman texting

david-rosales:
@kevinrobinson I broke up the rendering of the button label text as you suggested but am still unsure of what the bug was. Can you clarify just to make sure that it has been fixed? Thanks

ui/src/message_popup/mobile_prototype/text_footer_component.jsx
```diff
@@ -0,0 +1,151 @@
...
+        </div>
+      </div>
+    );
+  },
+
+  renderButtonText(){
+    if(this.props.isReadyToMoveOn) return this.props.nextButtonLabel;
+    if(this.props.helpType === 'feedback' && this.state.initialMessageSent) return 'Revise response';
+    return `Respond to ${this.props.mainStudent !== undefined ? this.props.mainStudent.name : 'Class'}`;
+  }
```
kevinrobinson: @david-rosales This looks good to me!  And a good place where setting up tests will be helpful :)

Formatting the code from https://github.com/mit-teaching-systems-lab/threeflows/pull/83/commits/b019b831572b99e47bdc5ec447dd47f7e88a1a9a, the bug came up when the user is first shown the question in solution mode, before they've typed anything:

```
const mainButtonLabel = this.props.isReadyToMoveOn
  ? this.props.nextButtonLabel
  : this.props.helpType !== 'feedback'
    ? 'Respond to ' + (this.props.mainStudent !== undefined ? this.props.mainStudent.name : 'Class')
    : this.state.initialMessageSent
      ? 'Revise Response'
      : 'Save Response';
```

looks good now!

david-rosales:
Alright awesome, merging now.

PR #84, opened 2016-07-13
Update API functions to return Promises

kevinrobinson:
In order to use `Promise.all` to join separate server requests, this PR:
- adds in `superagent-promise`
- updates `api.js` so that all requests return promises while still supporting existing callback usage
- updates the eslint config to allow ES6 globals like `Promise`

### DIFF
--- a/ui/.eslintrc
+++ b/ui/.eslintrc
@@ -1,6 +1,7 @@
 {
   "env": {
-    "browser": true
+    "browser": true,
+    "es6": true
   },
   "parser": "babel-eslint",
   "parserOptions": {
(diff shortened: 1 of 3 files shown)


kevinrobinson:
@david-rosales This lets us directly express "make two requests in parallel and let me know when they're both done" like this:

```
Promise.all([
  Api.evaluationsQuery(),
  Api.evidenceQuery()
]).then(this.onDataReceived);
```

Can you review?

david-rosales:
@kevinrobinson Looks good, Approved.

PR #88, opened 2016-07-15
Remove localhost Rollbar reporting

kevinrobinson:
No more reporting from localhost.

screen shot 2016-07-15 at 3 43 42 pm


david-rosales:
@kevinrobinson Good idea, approved.

PR #89, opened 2016-07-15
Add Do Something route, experiment with embedding videochat

kevinrobinson:
This adds a new route for a "Do Something, Not Nothing" game.  There's nothing in there now but an experimental embedding of appear.in to see if it would work to mix videochat with a game experience.  appear.in only works in an embedded iframe over SSL, so merging this to see if it will work.

This also extends the `NavigationAppBar` so the styling can be changed, and fixes a bug where the `title` being passed would be ignored.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -21,6 +21,7 @@
   },
   "homepage": "https://github.com/mit-teaching-systems-lab/threeflows#readme",
   "dependencies": {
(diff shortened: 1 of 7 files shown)


kevinrobinson:
This is experimental, going to merge it just to see if this works over SSL. cc @david-rosales

PR #90, opened 2016-07-15
Fix bug in feedback mode with showing examples when question has none

kevinrobinson:
I noticed this trying some new questions without examples yet.  This adds another check so that the "revision" prompt is only shown if we're in feedback mode and there are actual examples to show.

The bug:
screen shot 2016-07-15 at 4 51 02 pm


kevinrobinson:
@david-rosales Can you review?  Squashing a 🐛

david-rosales:

PR #92, opened 2016-07-18
Script for migrating playtest data dump into main datastore

kevinrobinson:
This is for loading some evaluations done during playtesting on a separate instance back into the main deployment.

These were separate deployments to prevent any live interaction between "players" and "evaluators" but merging these back together helps look at data across scorers.


kevinrobinson:
selfie

PR #93, opened 2016-07-18
Add MessagePopUp-specific evaluation data in raw assessment page

kevinrobinson:
For easier reviewing of evaluations done during playtest.  Not attached to this as a direction, but it's useful for reviewing data right now.

screen shot 2016-07-18 at 10 32 04 am


kevinrobinson:
selfie, to start reviewing this data

PR #98, opened 2016-07-19
Swap evaluator and candidate pages, abstract querying to Api

kevinrobinson:
The initial candidate page was actually filtering by evaluator email, so this moves that to an evaluator page and adds a proper candidate page.

It also moves some querying and joining code into the `api` module, and renames the fields to be clearer on the write path.


kevinrobinson:
selfie to check these out for analysis

PR #100, opened 2016-07-21
Add prototype flow through providing audio responses

kevinrobinson:
#### What is this
This adds a new page to prototype the flow of using audio responses in Message PopUp.

The flow looks like this:
screen shot 2016-07-21 at 1 05 18 pm

screen shot 2016-07-21 at 1 05 22 pm

screen shot 2016-07-21 at 1 05 27 pm

screen shot 2016-07-21 at 1 05 31 pm

screen shot 2016-07-21 at 1 05 35 pm

#### How does audio recording work?
It forks AuO and makes some hacks to use it's audio recording capabilities without the visualization or editing bits (a proper refactor or a minimal implementation to capture audio would be better, but this allows prototyping immediately).  This is wrapped up in the `AudioCapture` component, and I'll open a conversation with STEP amigos about this.

The data is captured in the browser and sent to new endpoints in the server.  They write .wav files to S3, and read them back for serving with an authorization guard.  This is a new S3 bucket just for this purpose.

`.wav` files are hundreds of kb even for a few seconds, and upload latency affects the UX even on Desktop wifi so it'll be good to test the prototype experience to see if we need to optimize further.

#### How does the flow work?
The `AudioRecorderFlow` component has minimal presentation itself, but manages the state transitions through the flow and interactions with the server.  Callers can provide their own presentation for each state.


kevinrobinson:
The two additional commits clean up how files are stored in S3, so they're YYYYMMDD style and split into different folders based on environment and domain.

kevinrobinson:
@david-rosales Can you review this for the overall flow of the experience and the code as well?

david-rosales:
@kevinrobinson Awesome! I tried running it on my computer and finally got it working, mostly. I came across a few errors that might be completely caused by my own setup or might actually happen across all builds on all setups. I'm not sure though so if you can test for the errors I found, it would be great!

For some reason, after multiple tries of clicking record, waiting, and then clicking done, it gets the following error after recording for some time:

```
Uncaught TypeError: Cannot read property 'length' of null     AuO.js:788

beginAudioSave @ AuO.js:788
onSaveClicked @ VM6710:464
AuO.save @ VM6710:42
(anonymous function) @ audio_capture.jsx:14
```

I tried pinpointing when exactly this occurred but it changed often. However, I did generally get the error when the recording was longer than ~6 seconds. It's possible that this is just the setup on my computer so can you please test it on yours?

Another thing that might be universal is that when clicking Record Again for the 6th time in a row, you get the following error:

```
Uncaught NotSupportedError: Failed to construct 'AudioContext':          VM10148:82
The number of hardware contexts provided (6) is greater than or 
equal to the maximum bound (6).

AuO @ VM10148:82
startRecording @ audio_capture.jsx:62
proxiedMethod @ VM9414:46
AudioCapture_componentDidUpdate @ audio_capture.jsx:53
AudioCapture_componentDidUpdate @ VM9414:46
invokeComponentDidUpdateWithTimer @ VM9531:68
...
```

This error is very consistent.

Another thing that might just be chrome, I'm not 100% sure is that when you record audio the red recording symbol shows up in the tab. However, this icon doesn't go away unless you refresh or manually redirect the page. It persists if you use the drawer to change directories. This might possibly mean that its still trying to record..? Or just that's how chrome works. 

Can you check if any of these errors are consistent with your setup?

kevinrobinson:
@david-rosales Thanks!  This is awesome feedback and a great bug report, many thanks!  I'll dig into this more when I'm back after next week, and see if this is something I can squash or start removing AuO to use the recording API directly.

kevinrobinson:
@david-rosales Thanks for the awesome testing and review!  This was coming from making a new Au instance every time `AudioCapture` changed state and started recording.  Instead each `AudioCapture`component now creates a single Au instance and re-uses it.  It's only when it fully unmounts that the resources are all freed.

kevinrobinson:
@david-rosales This is the essence of that change: https://github.com/mit-teaching-systems-lab/threeflows/pull/100/files#diff-fa0d926f4334fc3b93825c0107e68337R24

I manually tested this and verified it works, so I'm going to merge this now and then try to integrate this into the game as an actual response for one of the questions.

PR #101, opened 2016-07-21
Show indicators instead of competency groups when starting session

kevinrobinson:
This said "learning objectives" before when actually showing competency groups, so this clarifies and uses the indicator as the grain size shown to the player:

Before:
screen shot 2016-07-21 at 3 49 01 pm

After:
screen shot 2016-07-21 at 3 47 33 pm


kevinrobinson:
@david-rosales This came up from discussion after the playtest, could you review?

david-rosales:
@kevinrobinson approved!

PR #103, opened 2016-08-01
Fix flow errors for react-swipeable-views

kevinrobinson:
React released 0.15.2, which led to `npm install` on Travis pulling the new minor version.  This is good in general, but the point release includes a lot of new warnings, so this locks React at 0.15.1.0 and punts it for now to deal with this explicitly.


kevinrobinson:
selfie to unblock https://github.com/mit-teaching-systems-lab/threeflows/pull/101 on this unrelated dependency change

PR #54, opened 2016-06-29
Message PopUp: Prototype of mobile web scoring flow

kevinrobinson:
This is a functional prototype of a scoring flow for Message PopUp targeting mobile web.  It includes the work for this feature along with some related refactorings.

#### List of questions
This shows the list of questions that need to be scored.  Each question may have multiple responses from different students.  The icons are placeholder now, but could mirror what Joyce and Rupal come up with for different groups of competencies.
screen shot 2016-06-29 at 9 46 16 am

#### Scoring responses for a question
Tapping into a question shows the context about that question, including the question itself, card for the student, and any examples and non-examples.  These can be toggled open or closed.
screen shot 2016-06-29 at 12 12 08 pm

#### Swiping to scope
The intent of this prototype is for scorers to not feel overwhelmed and burdened by the amount of work, and to feel like they can tick down responses that need scoring fairly quickly.  Other learning experiences will certainly require more contemplation from faculty and evaluators, but this is an experiment with a quicker feedback format for this particular experience.

Swiping right scores it as "competency demonstrated," dismissing it and recording the score.
screen shot 2016-06-29 at 12 37 51 pm

And swiping left brings up a feedback note:
screen shot 2016-06-29 at 12 23 03 pm

#### Related changes and refactorings
- Introduces react-swipeable-views and react-motion for prototyping the swipe interaction
- Adds a new database table for holding evaluations
- Factors out API calls into an `api` module
- Adds in styling prop to `StudentCard` to support re-use in the scoring page
- Adds in question ids to the question data, and a workaround function for historical responses that don't have these ids

PR #104, opened 2016-08-02
Question Authoring

david-rosales:
I created the basic interface that will be the front of question authoring. The actual creation and deletion of questions does not exist yet because the questions need to be accessed by database for that to happen. Below are the various parts of the question authoring page and details on any trouble I ran into while making it.

This is the main question authoring page and what I decided to include on the navigation drawer. You might notice that archived questions was given its own page and there is more info on that down below. The basis behind the look of the page was the design on #54.
<div style="display: flex; flexDirection: horizontal">
screen shot 2016-07-29 at 4 13 55 pm
screen shot 2016-07-29 at 4 13 19 pm
</div>

Below is a set of 4 images that all show the various parts of the question editing page. All are as we discussed except the Good/Bad Examples are not collapsible. One problem I came across was that we did not discuss how to display the competency group or indicators, both essential to a question. I decided to go with a menu type component for the competency group that showed the competency name when selecting and the key when selected. It also displays the text behind the competency group above the selection. The indicators were trickier to display because there was no short version of it. It was either the id which held no meaning or the entirety of the indicator text. This did not work well with a menu component so for now I decided to use radio buttons. If there are any other better components we can use to display indicators please let me know.

<div style="display: flex; flexDirection: horizontal">
screen shot 2016-07-29 at 4 14 14 pm
screen shot 2016-07-29 at 4 14 38 pm
screen shot 2016-07-29 at 4 16 15 pm
screen shot 2016-07-29 at 4 15 14 pm
</div>

The bottom three images display the New Question page. It looks almost exactly the same as the edit question minus the "Original Question Text" which doesn't exist on the page. Another difference is the buttons at the bottom. In order for the "Create" button to become enabled, the question text textfield needs to be populated.
<div style="display: flex; flexDirection: horizontal">
screen shot 2016-07-29 at 4 15 32 pm
screen shot 2016-07-29 at 4 15 43 pm
screen shot 2016-07-29 at 4 16 15 pm
</div>

Below is an example of the search bar in action. It filters through the question text, learning objective, indicator, student names, student ids, question id, good examples, and bad examples. You can see the "Archived Questions" section at the bottom of the image and more information about it will be with the next image.
screen shot 2016-07-29 at 4 16 43 pm

One issue I came across while implementing the "Archived Questions" section was how to make it look distinguishable from all the other questions. I felt that having them on the same page, with the possibility of there being plenty of both questions and archived questions, would be too crowded and a bit hard to get to since it was at the very bottom of the page. It wouldn't be hard to see since the background was a different color but it would be a hassle to try and get to the archived questions every time. So I created an alternate page as well (so we can see which one works better) that displays only archived questions and has more options for them.
<div style="display: flex; flexDirection: horizontal">
screen shot 2016-07-29 at 4 17 11 pm
screen shot 2016-07-29 at 4 17 25 pm
screen shot 2016-07-29 at 4 17 37 pm
</div>

Because the archived questions then had their own page, I thought it would be useful to give users more options to deal with the archived questions including a restore and a permanent delete. The reason we archived the questions in the first place was so that the questions would not be completely deleted and older responses could still display the question information of when they were answered. I kept this in mind and made sure that it was reiterated when confirming a restore or a delete.

Below is an image of what happens when a user interacts with the "Archived Questions Page"
screen shot 2016-07-29 at 4 17 57 pm

One thing I was able to experiment with is the mass selection of archived questions to manage. The selection mode is enabled by pressing the Check icon at the top right and disabled by pressing the Close icon that replaces it once enabled. This can be seen below. 
<div style="display: flex; flexDirection: horizontal">
screen shot 2016-07-29 at 4 18 16 pm
screen shot 2016-07-29 at 4 18 33 pm
screen shot 2016-07-29 at 4 18 44 pm
</div>

Pressing on the RESTORE or DELETE buttons either on the selection mode or on the individual question will show the following dialog with the respective number of questions displayed.
<div style="display: flex; flexDirection: horizontal">
screen shot 2016-07-29 at 4 18 57 pm
screen shot 2016-07-29 at 4 19 11 pm
screen shot 2016-07-29 at 4 19 33 pm
screen shot 2016-07-29 at 4 19 45 pm
</div>

Below are a few examples of the search bar function which works the same as the one in the "Questions Page". Something to note in the third image below is that selections made persist even if the question is not displayed due to the search field. The image displays "3 selected" but the page only shows 1. This is because the other 2 questions that are selected did not pass the filter and are not being currently shown. Removing the search text will reshow everything as it was.
<div style="display: flex; flexDirection: horizontal">
screen shot 2016-07-29 at 4 20 16 pm
screen shot 2016-07-29 at 4 20 31 pm
screen shot 2016-07-29 at 4 21 03 pm
</div>

Those were the changes made that I thought were worth sharing. A few next steps I can think of are:
- Figuring out what to do with the competency groups. Since they are complex and have various parts should there also be an authoring section for them?
- Figuring out what to do with the archived questions. Is the new page a good idea or should we stick with displaying it on the same page as the normal questions.
- Moving questions over to a database. We should still keep the questions.js page around though for testing purposes without connection to a database.
- Actually creating, archiving, deleting, and restoring. The previous 4 actions are all displayed solely in the console and don't actually do anything right now. In order to do them, the questions first need to be on a database.


kevinrobinson:
@david-rosales Here are the notes from our earlier in-person conversation, broken down into two pull requests and some future improvements we could make but will punt for now:

```
learning objective and indicator, pull out learning objective
menus: group in message popup
archived questions
  add archived to main page
  click archived to see question via popup and restore
  remove archived page and menu
...
database
  add table with read and write endpoints
  load questions from endpoint
  on actions, call setState and then call save
  onCreate calls up to the page component to setState and save
...
punt: could do autocomplete for indicator later
punt: mobile web - layout for desktop
```

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -0,0 +1,162 @@
...
+        return questionString.toLowerCase().includes(value);
+      });
+      questions = questions.sort(questionOriginal => {
+        const question = withStudents([withLearningObjectiveAndIndicator(questionOriginal)])[0];
+        var questionString = question.id + " " + question.text + " " + question.learningObjective.key + " " + question.learningObjective.text + " " + question.learningObjective.competencyGroup + " " + question.indicator.text;
+        if(_.has(question, 'students')) _.forEach(question.students, student => questionString += " " + student.id + " " + student.name);
+        _.forEach(question.examples, example => questionString += " " + example);
+        _.forEach(question.nonExamples, example => questionString += " " + example);
+        return questionString.split(value).length;
+      });
```
kevinrobinson: I don't quite follow what the sort is doing here, can you explain?

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -0,0 +1,162 @@
...
+    }else{
+      var questions = _.clone(allQuestions);
+      questions = questions.filter(questionOriginal => {
+        const question = withStudents([withLearningObjectiveAndIndicator(questionOriginal)])[0];
+        var questionString = question.id + " " + question.text + " " + question.learningObjective.key + " " + question.learningObjective.text + " " + question.learningObjective.competencyGroup + " " + question.indicator.text;
+        if(_.has(question, 'students')) _.forEach(question.students, student => questionString += " " + student.id + " " + student.name);
+        _.forEach(question.examples, example => questionString += " " + example);
+        _.forEach(question.nonExamples, example => questionString += " " + example);
+        return questionString.toLowerCase().includes(value);
+      });
```
kevinrobinson: I'd split this out to a function, something like `searchTextForQuestion`.

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -0,0 +1,162 @@
...
+
+  onArchivedQuestions(){
+    Routes.navigate(Routes.messagePopupAuthorArchivedQuestionsPath());
+  },
+
+  onSearchBarChange(event){
+    const value = event.target.value.toLowerCase().trim();
+    if(value === ''){
+      this.setState({questions: allQuestions});
+    }else{
```
kevinrobinson: As a style suggestion, I'd recommend doing the check for the empty string and then calling setState and returning.  Then the rest of the method can be un-indented and you can remove the `else`.

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -0,0 +1,162 @@
...
+  onNewQuestion(){
+    Routes.navigate(Routes.messagePopupAuthorQuestionsNewPath());
+  },
+
+  onArchivedQuestions(){
+    Routes.navigate(Routes.messagePopupAuthorArchivedQuestionsPath());
+  },
+
+  onSearchBarChange(event){
+    const value = event.target.value.toLowerCase().trim();
```
kevinrobinson: Consider what happens when the user enters search text with a space, which I think might be common in this case.  What would the user expect to happen?

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -0,0 +1,162 @@
...
+        _.forEach(question.nonExamples, example => questionString += " " + example);
+        return questionString.split(value).length;
+      });
+      this.setState({questions});
+    }
+  },
+
+  render(){
+    var archivedQuestions = [];
+    archivedQuestions.push({
```
kevinrobinson: Even for testing locally, please split test data outside of the `render` method into another file or at least a constant at the top of the component.  Having data like this inside `render` can only lead to accidentally merging it :)

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -0,0 +1,162 @@
...
+    top: 112,
+    bottom: 0,
+    overflowY: 'scroll'
+  },
+  currentQuestionsContainer: {
+    marginBottom: 10
+  },
+  archivedQuestionsContainer: {
+    marginBottom: 10,
+    backgroundColor: "#F4F4F4"
```
kevinrobinson: Can you use a [Material UI color](http://www.material-ui.com/#/customization/colors) here?

ui/src/message_popup/author/question_button.jsx
```diff
@@ -0,0 +1,34 @@
...
+import SchoolIcon from 'material-ui/svg-icons/social/school';
+
+export default React.createClass({
+  displayName: 'QuestionButton',
+
+  propTypes: {
+    question: React.PropTypes.object.isRequired
+  }, 
+
+  changePath(){
```
kevinrobinson: I think I probably missed this earlier, but please use the convention where event handlers start with `on` and indicate what event occurred, so something like `onItemClicked`.

ui/src/message_popup/author/question_button.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+
+import {withLearningObjectiveAndIndicator} from '../transformations.jsx';
+import * as Routes from '../../routes.js';
+
+import {ListItem} from 'material-ui/List';
+
+import SchoolIcon from 'material-ui/svg-icons/social/school';
+
+export default React.createClass({
```
kevinrobinson: nice factoring!

ui/src/app.jsx
```diff
@@ -39,6 +39,10 @@ export default React.createClass({
     '/message_popup/exploration': 'messagePopupExploration',
     '/message_popup/evaluations/:id': 'messagePopupEvaluation',
     '/message_popup/scoring': 'messagePopupScoring',
+    '/message_popup/author/questions' : 'messagePopupAuthorQuestions',
+    '/message_popup/author/questions/new' : 'messagePopupAuthorQuestionsNew',
+    '/message_popup/author/questions/:id' : 'messagePopupAuthorQuestionsEdit',
```
kevinrobinson: Nice!

ui/src/message_popup/author/question_page.jsx
```diff
@@ -0,0 +1,284 @@
...
+import Dialog from 'material-ui/Dialog';
+import FlatButton from 'material-ui/FlatButton';
+
+import ArrowBackIcon from 'material-ui/svg-icons/navigation/arrow-back';
+
+export default React.createClass({
+  displayName: 'QuestionPage',
+
+  propTypes: {
+    originalQuestion: React.PropTypes.object,
```
kevinrobinson: Would it make sense to just use `originalQuestion` and remove the other props here?  That seems like the essential difference between `new` and `edit`, and then `getInitialState` could just read from the original question if there was one or fall back to the default.

That would also allow the code to form `goodExamplesText` and `badExamplesText` to be moved here (where it seems like it really belongs, since the newline-delimited format is specific to this UI).

ui/src/message_popup/author/question_page.jsx
```diff
@@ -0,0 +1,284 @@
...
+      students: this.props.students,
+      learningObjective: this.props.learningObjective,
+      indicator: this.props.indicator,
+      goodExamplesText: this.props.goodExamplesText,
+      badExamplesText: this.props.badExamplesText,
+      deleteConfirmationOpen: false
+    });
+  },
+
+  returnToQuestions(){
```
kevinrobinson: I left this a few places, but could you do a pass and update the event handlers to have names that indicate what happened, something like `onReturnToQuestionsClicked`?

ui/src/message_popup/author/question_page.jsx
```diff
@@ -0,0 +1,284 @@
...
+    const alreadyAdded =  student !== undefined && _.find(this.state.students, newStudent => newStudent.id === student.id) !== undefined;
+    if(student !== undefined && !alreadyAdded){
+      this.setState({students: this.state.students.concat(student)});
+    }
+  },
+
+  removeStudent(studentId){
+    return function(){
+      this.setState({students: _.remove(_.clone(this.state.students), student => student.id !== studentId)});
+    }.bind(this);
```
kevinrobinson: Do you need this bind here?

ui/src/message_popup/author/question_page.jsx
```diff
@@ -0,0 +1,284 @@
...
+    const student = _.find(allStudents, student => student.name.toLowerCase() === studentName.toLowerCase());
+    const alreadyAdded =  student !== undefined && _.find(this.state.students, newStudent => newStudent.id === student.id) !== undefined;
+    if(student !== undefined && !alreadyAdded){
+      this.setState({students: this.state.students.concat(student)});
+    }
+  },
+
+  removeStudent(studentId){
+    return function(){
+      this.setState({students: _.remove(_.clone(this.state.students), student => student.id !== studentId)});
```
kevinrobinson: Use `filter` or `_.without` and you don't need to clone, they won't mutate the underlying array like `remove` does.

ui/src/message_popup/author/question_page.jsx
```diff
@@ -0,0 +1,284 @@
...
+  },
+
+  getExamples(){
+    const {goodExamplesText, badExamplesText} = this.state;
+    var goodExamples = goodExamplesText.split('\n\n');
+    var badExamples = badExamplesText.split('\n\n');
+    goodExamples = goodExamples.map(example => example.trim());
+    badExamples = badExamples.map(example => example.trim());
+    _.remove(goodExamples, example => example == '');
+    _.remove(badExamples, example => example == '');
```
kevinrobinson: This seems like a candidate for a method that you could use for both good and bad examples, something like `parseExampleText` that takes a string and does this transformation?

ui/src/message_popup/author/question_editing_components/learning_objectives.jsx
```diff
@@ -0,0 +1,54 @@
...
+import Divider from 'material-ui/Divider';
+import DropDownMenu from 'material-ui/DropDownMenu';
+import MenuItem from 'material-ui/MenuItem';
+
+export default React.createClass({
+  displayName: 'LearningObjectives',
+
+  propTypes: {
+    learningObjective: React.PropTypes.object.isRequired,
+    onChangeLearningObjective: React.PropTypes.func.isRequired
```
kevinrobinson: Since this method is all about the learning objective changing, I'd suggest having this component handle the event from `DropDownMenu`, and then only calling `this.props.onChangeLearningObjective` with the `learningObjectiveId`.  That way the reference to the DOM event and other things that aren't important to the caller of this component are hidden.

ui/src/message_popup/author/question_editing_components/indicators.jsx
```diff
@@ -0,0 +1,47 @@
...
+import Paper from 'material-ui/Paper';
+import Divider from 'material-ui/Divider';
+import {RadioButton, RadioButtonGroup} from 'material-ui/RadioButton';
+
+export default React.createClass({
+  displayName: 'Indicators',
+
+  propTypes: {
+    indicator: React.PropTypes.object.isRequired,
+    onChangeIndicator: React.PropTypes.func.isRequired
```
kevinrobinson: Let's update to `onIndicatorChanged`, and I'll add this to a style guide.

ui/src/message_popup/author/edit_question_page.jsx
```diff
@@ -0,0 +1,35 @@
...
+  displayName: 'EditQuestionPage',
+
+  propTypes: {
+    questionId: React.PropTypes.string.isRequired
+  },
+
+  render(){
+    const question = withLearningObjectiveAndIndicator(_.find(withStudents(allQuestions), question => question.id.toString()===this.props.questionId));
+    var goodExamplesText = '';
+    _.forEach(question.examples, example => {goodExamplesText += (goodExamplesText === '' ? '' : '\n\n') + example;});
```
kevinrobinson: Prefer `const` over `var` wherever you can, since it provides a guarantee that the value won't be changed later accidentally.  In this case would this work?

```
const goodExamplesText = question.examples.join('\n');
```

ui/src/message_popup/author/archived_questions.js
```diff
@@ -0,0 +1,150 @@
+/* @flow weak */
+export const allArchivedQuestions = [
+  {
```
kevinrobinson: 👍  great way to start developing this!

ui/src/message_popup/author/archived_question_button.jsx
```diff
@@ -0,0 +1,61 @@
...
+    return (
+      <div>
+        {!this.props.checking && 
+          <ListItem
+            style={{cursor: 'pointer', fontSize: 14}}
+            leftIcon={<SchoolIcon />}
+            onTouchTap={function(){
+              this.onTouch();
+              this.props.onSetDialog();
+            }.bind(this)}
```
kevinrobinson: A good heuristic is that if handlers are multi-line, pull them out into their own method.

ui/src/message_popup/author/archived_questions_page.jsx
```diff
@@ -0,0 +1,272 @@
...
+import CheckedIcon from 'material-ui/svg-icons/toggle/check-box';
+import SchoolIcon from 'material-ui/svg-icons/social/school';
+
+import NavigationAppBar from '../../components/navigation_app_bar.jsx';
+import ArchivedQuestionButton from './archived_question_button.jsx';
+
+import {withStudents, withLearningObjectiveAndIndicator} from '../transformations.jsx';
+import {allArchivedQuestions} from './archived_questions.js';
+
+export default React.createClass({
```
kevinrobinson: This is a good bit of code we can delete to simplify! :)

ui/src/message_popup/author/question_editing_components/editing_component.jsx
```diff
@@ -0,0 +1,14 @@
+/* @flow weak */
+import QuestionText from './question_text.jsx';
+import Students from './students.jsx';
+import Examples from './examples.jsx';
+import LearningObjectives from './learning_objectives.jsx';
+import Indicators from './indicators.jsx';
+
+export {
```
kevinrobinson: I imagine you're following the pattern in the Message PopUp page, which is a good strategy!  But this doesn't seem particularly useful to me, you can just import these directly.

The file exporting specific pieces in `message_popup/index.js` is because I was trying to be explicit about what the entry points were into the MessagePopUp UI from other parts of the product (e.g, viewing scores across all products, navigating from the shared navigation drawer).  As we were talking about in-person, I haven't done enough to talk about this or make this clear enough.  But you don't need to worry about that here, and can just import these editing components directly in the question page.

ui/src/message_popup/author/question_page.jsx
```diff
@@ -0,0 +1,284 @@
...
+  },
+
+  onGoodExamplesChange(event, text){
+    this.setState({goodExamplesText: text});
+    this.getExamples();
+  },
+
+  onBadExamplesChange(event, text){
+    this.setState({badExamplesText: text});
+    this.getExamples();
```
kevinrobinson: Does calling `getExamples` here or in `onGoodExamplesChange` do anything?

ui/src/message_popup/author/question_editing_components/learning_objectives.jsx
kevinrobinson: Err sorry scratch that I forget we were removing this :)

kevinrobinson: Actually the same idea applies to `onChangeIndicator` for the indicator component, so double 😄

ui/src/message_popup/author/question_page.jsx
```diff
@@ -0,0 +1,284 @@
...
+            <EditingComponents.LearningObjectives 
+              learningObjective={this.state.learningObjective}
+              onChangeLearningObjective={this.onChangeLearningObjective}
+              />
+            <EditingComponents.Indicators 
+              indicator={this.state.indicator}
+              onChangeIndicator={this.onChangeIndicator}
+              />
+
+            <div style={styles.finalButtonRow}>
```
kevinrobinson: It's up to you, but there might be enough in these buttons for it to be worth splitting into its own method.

ui/src/message_popup/author/question_page.jsx
```diff
@@ -0,0 +1,284 @@
...
+              <FlatButton
+                label="Cancel"
+                onTouchTap={this.closeDeleteConfirmation}/>,
+              <FlatButton
+                label="Delete"
+                style={{color: "#CA2300"}}
+                onTouchTap={function(){
+                  this.closeDeleteConfirmation();
+                  this.onDeleteQuestion();
+                }.bind(this)}/>
```
kevinrobinson: I'd recommend splitting this out to a method, it's multiple lines and then you can remove the `bind` as well.  Also, if you use the ES6 arrow syntax for functions, `this` will be bound for you.

ui/src/message_popup/author/question_page.jsx
```diff
@@ -0,0 +1,284 @@
...
+
+  onChangeLearningObjective(event, index, value){
+    this.setState({learningObjective: _.find(learningObjectives, learningObjective => learningObjective.id === value)});
+  },
+
+  onChangeIndicator(event, value){
+    this.setState({indicator: _.find(indicators, indicator => indicator.id.toString() === value)});
+  },
+
+  saveCheckPassed(){
```
kevinrobinson: nice way to wrap this check up!

ui/src/message_popup/author/question_page.jsx
```diff
@@ -0,0 +1,284 @@
...
+  },
+
+  onQuestionTextChange(event, value){
+    this.setState({questionText: value});
+  },
+
+  addStudent(studentName){
+    const student = _.find(allStudents, student => student.name.toLowerCase() === studentName.toLowerCase());
+    const alreadyAdded =  student !== undefined && _.find(this.state.students, newStudent => newStudent.id === student.id) !== undefined;
+    if(student !== undefined && !alreadyAdded){
```
kevinrobinson: Would it work to filter the list of students so that students already added couldn't be added again, rather than having to guard for that here?

ui/src/message_popup/author/question_page.jsx
```diff
@@ -0,0 +1,284 @@
...
+      text,
+      examples: goodExamples,
+      nonExamples: badExamples,
+      indicatorId: this.state.indicator.id,
+      indicator: this.state.indicator,
+      learningObjectiveId: this.state.learningObjective.id,
+      learningObjective: this.state.learningObjective
+    };
+    console.log(question);
+    this.returnToQuestions();
```
kevinrobinson: These look good!  For `onCreated`, `onDeleted` and `onUpdated`, we can update those in the next PR.

kevinrobinson:
@david-rosales This is awesome work!  I'd suggest working on the design changes first, and then any comments here and let me know if you want to talk out anything else here or if I can help with a final review.

david-rosales:
@kevinrobinson I removed `ArchivedQuestionsPage`, references to learningObjectives, and made changes based on you comments. Can you review it when you get a chance?

ui/src/message_popup/author/archived_questions.js
```diff
@@ -0,0 +1,150 @@
...
+      'How \'bout an example??',
+      'How \'bout an example???',
+      'How \'bout an example????',
+      'How \'bout an example?????',
+    ],
+    nonExamples: [
+      'How \'bout a bad example?'
+    ],
+    indicatorId: 501,
+    learningObjectiveId: 50
```
kevinrobinson: You can cut out the `learningObjectiveId` fields here too.

ui/src/message_popup/author/edit_question_page.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+export default React.createClass({
+  displayName: 'EditQuestionPage',
+
+  propTypes: {
+    questionId: React.PropTypes.string.isRequired
+  },
+
+  render(){
+    const question = withIndicator(_.find(withStudents(allQuestions), question => question.id.toString()===this.props.questionId));
```
kevinrobinson: Would it make sense to lookup the question first, then join in the student and indicator data?

ui/src/message_popup/author/new_question_page.jsx
```diff
@@ -0,0 +1,14 @@
+/* @flow weak */
+import React from 'react';
+
+import QuestionPage from './question_page.jsx';
```
kevinrobinson: This component looks like it doesn't do anything, would it make sense to just call `QuestionPage` directly?

kevinrobinson:
@david-rosales This looks good, approved!

Let me know if you want to pair on the database bits this afternoon or Monday morning, otherwise I probably won't be able to do it until Tuesday afternoon.

david-rosales:
@kevinrobinson Ok, I'm merging now. I can pair with you on the database things Monday morning. Want to schedule that?

kevinrobinson:
@david-rosales awesome, done!

PR #105, opened 2016-08-02
Add July updates for virtual school students

kevinrobinson:
Adds updates to the virtual school after some work from Eliza.  No learning experiences use this yet.

Screenshot:
screen shot 2016-08-02 at 5 05 40 pm

Navigation:
screen shot 2016-08-02 at 5 05 51 pm

### DIFF
--- a/ui/src/data/virtual_school.js
+++ b/ui/src/data/virtual_school.js
@@ -376,3 +376,6 @@ export const allStudents = initialStudents.concat(
   julyStudentsFromMursion,
   julyStudentsNotFromMursion
 );
+
+// Only include the latest cohort of students (others are essentially archived).
+export const activeStudents = allStudents.filter(student => student.cohortKey === '20160722');
--- a/ui/src/data/virtual_school.js
+++ b/ui/src/data/virtual_school.js
@@ -1,5 +1,37 @@
 /* @flow weak */
-export const allStudents = [
+
+/*
+This script can be used to read data from a Google Docs table.
+var text = `
+`;
+var pieces = text.trim().split('\n\n').join('\n').split('\n');
+function orNull(value) { return value === '' ? null : value; }
+console.log(JSON.stringify({
+  id: null,
+  grade: `7th grade`,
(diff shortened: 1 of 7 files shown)


kevinrobinson:
@david-rosales Can you review por favor?

ui/src/data/virtual_school.js
```diff
@@ -135,4 +167,212 @@ export const allStudents = [
...
+  }
+].map(addFields({
+  isMursion: true,
+  cohortKey: '20160722'
+}));
+
+export const allStudents = initialStudents.concat(
+  julyStudentsFromMursion,
+  julyStudentsNotFromMursion
+);
```
david-rosales: For now it shouldn't be a problem but don't some of the students from `julyStudentsNotFromMursion` overlap with `initialStudents`? The entire list of students isn't really shown anywhere right now so it doesn't really matter but if I remember correctly, the student autocomplete in #104 uses `allStudents`. This would lead to having, as an example, both Mike and Mike Bilson who are the same except for the addition of an academic performance field and a change to the family background field. Should we update the original ones instead or just leave this as is until we are ready to use them?

david-rosales:
@kevinrobinson I made one comment but it doesn't really pose too big of a problem since the only issue is present in an authoring tool that is not yet complete. Approved.

ui/src/data/virtual_school.js
kevinrobinson: Yeah, this is a great point!  In general, I think the idea is that the list of students is generally immutable, and that rather than changing students we create new variants of them.  This helps make clear that we look back at data from 7/12 that says "this person said X for student Y" that the meaning of student Y doesn't change after the fact.

In https://github.com/mit-teaching-systems-lab/threeflows/pull/105/files#diff-72bb72db1d42868cb18fa1c57dbf268eR21 it filters for only the latest cohort, so to make it simpler for #104 I moved that into `virtual_school.js` so that you can pull from activeStudents as well and get the desired autocomplete.

Thanks for the great feedback!

PR #106, opened 2016-08-02
Make no-unused-vars lint rule stricter

kevinrobinson:
So that any unused vars in function arguments are lint errors.  In some places, functions were specifying what values they would be called with, which is helpful but then the linter can't warn against arguments that are never used in the function body.


kevinrobinson:
@david-rosales Seem reasonable?

david-rosales:
@kevinrobinson Looks good!

PR #107, opened 2016-08-02
Remove references to learning objectives from Message PopUp

kevinrobinson:
An earlier pull request changed this to use the finer-grained indicators that the scenarios are aligned to (rather than the larger-grained WWA learning objectives).

This removes references to learning objectives and competencies altogether, in favor of the finer-grained indicators.  Aggregating indicators into views of proficiency in learning objectives, competencies, or other higher-level views will be done separately.


kevinrobinson:
@david-rosales This is the cleanup we talked about earlier today.  It may introduce some small merge conflicts with your open PRs but those should be straightforward to resolve.  Can you review?

ui/src/message_popup/transformations.jsx
```diff
@@ -13,13 +12,11 @@ export function withStudents(questions) {
...
-  const learningObjective = _.find(learningObjectives, { id: question.learningObjectiveId });
+export function withIndicator(question) {
   const indicator = _.find(indicators, { id: question.indicatorId });
   return {
     ...question,
-    indicator,
-    learningObjective
+    indicator
   };
 }
```
david-rosales: Now that this exists, can the inside of `questionsForIndicator` be replaced with `withIndicator`?

david-rosales:
@kevinrobinson I made a small suggestion which would remove some redundant code. Other than that, approved.

ui/src/message_popup/transformations.jsx
kevinrobinson: Great suggestion, thanks!  I incorporated this and also discovered a bug in making that change, mind re-reviewing?  I'll call out where it is.

ui/src/message_popup/scaffolding_card.jsx
```diff
@@ -46,21 +46,21 @@ export default React.createClass({
   getQuestions(selectedIndicatorId){
     var {isSolutionMode} = this.state;
     const questions = (isSolutionMode || selectedIndicatorId === ALL_INDICATORS)
-      ? _.shuffle(withStudents(allQuestions))
```
kevinrobinson: @david-rosales I noticed that the old `questionsForIndicator` had `_.shuffle` in it, which I wasn't expecting and so I went to factor it out since both branches here were shuffling.  But then I realized `getQuestions` sounds like it should be idempotent and return the same value every time you call it with the same state and props, but shuffle meant it wasn't.  So I updated this code to remove the shuffling and put it only at the point where this component reports the new session configuration, in `onSave`.

Does this make sense and look good?  Thanks for re-reviewing this part!

ui/src/message_popup/scaffolding_card.jsx
```diff
@@ -119,7 +119,7 @@ export default React.createClass({
         {showSlider &&
           <div>
             <div style={styles.optionTitle}>Session Length: {sessionLength} {sessionLength===1 ? "question" : "questions"}</div>
-            <Slider key={selectedIndicatorId} value={sessionLength} min={1} max={questions.length} step={1} onChange={this.onSliderChange}/>
```
david-rosales: The reasoning behind having the key set as the `selectedIndicatorId` was so that the slider would be forced to update when the indicator got changed. For some reason, updating the max wont be seen until the next setState. By setting a key that changes, it essentially created a new slider for each indicator which got around the issue of the max not updating. It's possible that the slider not updating is a problem caused elsewhere though.

david-rosales:
@kevinrobinson Alright, I found one thing that does cause an issue with the instructions page but everything else looks good.

ui/src/message_popup/scaffolding_card.jsx
kevinrobinson: ah good, thanks for explaining!  That's a good way to use `key` to workaround, but also clearly a bug in the library with not updating based on the prop change.

I went and did the good open source citizen thing and submitted the fix upstream in https://github.com/callemall/material-ui/pull/4895, so hopefully they'll pull it in and in the meantime I'll add back your great solution.  Thanks!

ui/src/message_popup/scaffolding_card.jsx
kevinrobinson: @david-rosales Ping on this, mind re-reviewing this part when you get a chance?  Thanks!

david-rosales:
@kevinrobinson Looks good!

PR #108, opened 2016-08-02
Add lint rule for React method ordering, and update code to match

kevinrobinson:
This adds a new eslint rule for the order of React methods.  It's roughly:

```
lifecycle
(anything else, eg., computing values from state or performing effects like navigating pages)
event handlers, prefixed with on
render
other rendering methods
```

### DIFF
--- a/ui/.eslintrc
+++ b/ui/.eslintrc
@@ -21,10 +21,28 @@
     "indent": ["error", 2],
     "semi": ["error", "always"],
     "comma-dangle": ["error", "only-multiline"],
-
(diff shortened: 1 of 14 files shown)


kevinrobinson:
@david-rosales This came out of the pull requests earlier today to help with remember these things.  I was going to add a style guide but found this awesome linting rule that can help check automatically.  Can you review, does this seem like a good idea?

david-rosales:
@kevinrobinson Ooo this looks cool and would definitely help with structuring, approved!

PR #109, opened 2016-08-04
Changed encodedText to directly include message structure

david-rosales:
Instead of parsing through the `encodedText` string, the code now directly reads the message data. Although it takes more time to create the actual questions, in the future it should not be an issue since it will be done through an authoring tool and not the code.

Instead of `encodedText` looking like: 
```
encodedText: '[:i]Some info text here.[:8s]Some student text from Ada.[:u]Some user text.',
```

It now looks like:
```
encodedText: [
  {
    type: 'info',
    text: 'Some info text here.'
  },
  {
    type: 'student',
    studentId: 8,
    text: 'Some student text from Ada.'
  },
  {
    type: 'user',
    text: 'Some user text.'
  }
],
```

### DIFF
--- a/ui/src/message_popup/mobile_prototype/mobile_interface.jsx
+++ b/ui/src/message_popup/mobile_prototype/mobile_interface.jsx
@@ -62,30 +62,23 @@ export default React.createClass({
   },
 
   getInitialMessages(questionObject){
-    var messages = [];
-    if(_.has(questionObject, 'encodedText')){
-
-      for (var index in questionObject.encodedText){
-        const textObject = questionObject.encodedText[index];
-        const student = _.find(this.props.question.students, student => student.id === textObject.studentId);
-        const message = {
-          type: textObject.type,
-          text: textObject.text,
-          key: textObject.text,
-          student
-        };
-        messages.push(message);
-      }
-    }else{
-      const message = {
+    if(_.has(questionObject, 'messages')) return this.props.question.messages.map(message => { 
+      return ({
+        type: message.type,
+        text: message.text,
+        key: message.text,
+        student: _.find(this.props.question.students, {id: message.studentId})
+      }); 
+    });
+    
+    return [
+      {
         type: 'info',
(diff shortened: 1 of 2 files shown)


david-rosales:
@kevinrobinson can you review this?

ui/src/message_popup/questions.js
```diff
@@ -34,7 +34,21 @@ export const inquiryQuestions:[Question] = [
     studentIds: [4], 
     id: 101,
     text: `At the conclusion of your lesson plan for this challenge, you seed a group discussion by asking "What are you curious about related to photosynthesis?"  Hayin says "Why are plants green?"  What do you do?`,
-    encodedText: `[:i]At the conclusion of your lesson plan for this challenge, you seed a group discussion.[:u]What are you curious about related to photosynthesis?[:4s]Why are plants green?`,
+    encodedText: [
```
kevinrobinson: Would `messages` be more direct than `encodedText`, now that the encoding is removed?

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -64,29 +64,17 @@ export default React.createClass({
...
-            text: messageText,
-            key: messageText,
-            student
-          };
-          messages.push(message);
-        }
+
+      for (var index in questionObject.encodedText){
+        const textObject = questionObject.encodedText[index];
+        const student = _.find(this.props.question.students, student => student.id === textObject.studentId);
```
kevinrobinson: A few style suggestions for you.

For transforming a list of values, I'd suggest using `.map` over iteration with for loops in general.
```
const messages = question.messages.map((message) => {
  ...
});
```

Also you can use a lodash shorthand for find here that might be more direct:

```
const student = _.find(this.props.question.students, { id: message.studentId});
```

kevinrobinson:
@david-rosales This is a good simplification, approved!  Just small style suggestions.

PR #110, opened 2016-08-04
Fixing shrinking icons

david-rosales:
I noticed that the icon on the right (user icon) was now shrinking after we simplified some of the flexbox properties.

<div style="display: flex;">
screen shot 2016-08-04 at 3 23 28 pm
screen shot 2016-08-04 at 3 26 16 pm
</div>

I added in `flexShrink: 0` to `styles.messageIcon` to prevent any icons from shrinking. I also set  `height: 0` in `styles.messageIconButton` to get rid of the additional padding created by the `IconButton`. Lastly, `marginTop: 5` was also added to `styles.messageIcon` in order to align both the left and right icons in terms of how high next to the message bubble they appeared. Before, the left icons were slightly shorter than the right icons.

screen shot 2016-08-04 at 3 32 07 pm

### DIFF
--- a/ui/src/message_popup/mobile_prototype/message_components.jsx
+++ b/ui/src/message_popup/mobile_prototype/message_components.jsx
@@ -43,7 +43,7 @@ const Message = React.createClass({
           <Paper style={messageTextStyle}>{this.props.text}</Paper>
           
         </div>
-        {this.props.rightIcon}
+        <div>{this.props.rightIcon}</div>
       </div>
     );
   }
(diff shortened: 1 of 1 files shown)


david-rosales:
@kevinrobinson This is an extension to what we did yesterday because a new bug popped up. Can you review it?

kevinrobinson:
@david-rosales Good catch!  Check out how we solved this on the left side in https://github.com/kevinrobinson/threeflows/commit/4aae1005e0fb65de7be2906562130835473bc79c, could you use something similar to that for the right side?  It'd be great to avoid having `height: 0` or other properties that work but aren't expressing what we really want to have happen.

david-rosales:
@kevinrobinson On the left we fixed the size of the element (the IconButton) containing the Icon. This prevented the padding around the icon from growing. However, with the right, there is no element around the icon and the icon itself is shrinking. The width is already set to 30 but it still shrinks.

david-rosales:
@kevinrobinson This is what we discussed, merging now.

kevinrobinson:
do it! 👍

PR #111, opened 2016-08-05
Staggered message-like animations

david-rosales:
I ran into a few issues while trying to stagger the animations. First being that I couldn't figure out how to use stagger with VelocityComponent (which I am currently assuming can't be done because the components have no knowledge of each other). I decided to then switch over to a VelocityTransitionGroup which does support stagger but two problems arose:
- I could not differentiate animations from left and right. It had to be a single animation, either left or right sliding. This also was not easy to work around because the messages that should be animated from left and from right were intermixed and under the same div. I could not simply have multiple VelocityTransitionGroups. Creating multiple ones would also affect the staggering.
- VelocityTransitionGroup seems to kill the `scrollIntoView` that I've been using. I'm assuming this has to do with setting the display to none while it is still there and so nothing scrolls. This does not happen with VelocityComponents however.

After running into those roadblocks, I decided it would be easiest to handle the staggering myself with a setInterval when messages were added in groups (addMessages lent itself very easily for this). So the animations occur as soon the message is added but the message adding is what is staggered.

### DIFF
--- a/ui/src/message_popup/mobile_prototype/mobile_interface.jsx
+++ b/ui/src/message_popup/mobile_prototype/mobile_interface.jsx
@@ -44,21 +44,22 @@ export default React.createClass({
     const goodExamples = _.map(this.props.question.examples, (example) => { return {type: 'Good', text: example}; });
     const badExamples = _.map(this.props.question.nonExamples, (example) => { return {type: 'Bad', text: example}; });
     const allExamples = _.shuffle(_.clone(goodExamples).concat(badExamples)); 
-    const messages = this.getInitialMessages(this.props.question);
     return ({
-      messages,
+      messages: [],
       allExamples,
       initialResponse: undefined,
       revisedResponse: undefined,
       elapsedMs: 0,
       exampleIndex: 0,
       dialog: null,
-      messageIndex: 1
+      messageIndex: 1,
+      disabledButton: false
     });
   },
 
   componentDidMount() {
     this.setInterval(this.updateTimer, ONE_SECOND);
+    this.addMessages(this.getInitialMessages(this.props.question));
   },
 
   getInitialMessages(questionObject){
@@ -66,7 +67,6 @@ export default React.createClass({
       return ({
         type: message.type,
         text: message.text,
-        key: message.text,
         student: _.find(this.props.question.students, {id: message.studentId})
       }); 
     });
@@ -75,7 +75,6 @@ export default React.createClass({
       {
         type: 'info',
         text: questionObject.text,
-        key: questionObject.text
       }
     ];
     
@@ -116,14 +115,17 @@ export default React.createClass({
   },
 
   addMessages(messageList){
-    var messages = _.clone(this.state.messages);
-    var messageId = this.state.messageIndex;
-    for(var messageIndex = 0; messageIndex < messageList.length; messageIndex++){
-      messageId+=1;
-      const message = _.extend({key: messageId}, messageList[messageIndex]);
-      messages.push(message);
-    }
-    this.setState({messages, messageIndex: messageId});
+    const message = {key: this.state.messageIndex+1, ...messageList.shift()};
+    this.setState({messages: this.state.messages.concat(message), messageIndex: this.state.messageIndex + 1, disabledButton: messageList.length > 0});
+    const interval = setInterval(function(){
+      if(messageList.length > 0){
+        const message = {key: this.state.messageIndex+1, ...messageList.shift()};
+        this.setState({messages: this.state.messages.concat(message), messageIndex: this.state.messageIndex + 1});
+      }else{
+        if(this.state.disabledButton) this.setState({disabledButton: false});
+        clearInterval(interval);
+      }
+    }.bind(this, interval, messageList), 300);
   },
 
   setInitialResponse(response){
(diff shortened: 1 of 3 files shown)


ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -44,29 +44,29 @@ export default React.createClass({
     const goodExamples = _.map(this.props.question.examples, (example) => { return {type: 'Good', text: example}; });
     const badExamples = _.map(this.props.question.nonExamples, (example) => { return {type: 'Bad', text: example}; });
     const allExamples = _.shuffle(_.clone(goodExamples).concat(badExamples)); 
-    const messages = this.getInitialMessages(this.props.question);
     return ({
-      messages,
+      messages: [],
```
david-rosales: I moved away from adding setting the InitialMessages here in getInitialState so that they could be added using the staggering from addMessages in componentDidMount.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -44,29 +44,29 @@ export default React.createClass({
...
       allExamples,
       initialResponse: undefined,
       revisedResponse: undefined,
       elapsedMs: 0,
       exampleIndex: 0,
       dialog: null,
-      messageIndex: 1
+      messageIndex: 1,
+      disabledButton: false
```
david-rosales: disabledButton was added so that animations could finish before sending in another message. Without this, users could beat the revision/example message with another message leading to two user messages before the revision/example message.

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -75,7 +75,6 @@ export default React.createClass({
       {
         type: 'info',
         text: questionObject.text,
-        key: questionObject.text
```
david-rosales: I removed the keys from getInitialMessages so that addMessages could handle it.

david-rosales:
@kevinrobinson Can you review this?

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -116,14 +115,17 @@ export default React.createClass({
...
   addMessages(messageList){
-    var messages = _.clone(this.state.messages);
-    var messageId = this.state.messageIndex;
-    for(var messageIndex = 0; messageIndex < messageList.length; messageIndex++){
-      messageId+=1;
-      const message = _.extend({key: messageId}, messageList[messageIndex]);
-      messages.push(message);
-    }
-    this.setState({messages, messageIndex: messageId});
+    const message = {key: this.state.messageIndex+1, ...messageList.shift()};
```
kevinrobinson: In general, it's best to avoid mutating objects like `messageList` that are passed to function unless that's explicitly called out or important for memory or performance reasons.  Modifying an argument can be particularly confusing when the method schedules future changes to the argument that will happen async.  It's difficult for the caller to know that those changes will occur or when, and if it mutates the value on its own that may not be what the code within `addMessages` expects.

I think the approach here is good - to schedule messages being added and then animate them individually, since `velocity-ui` doesn't directly support staggering with different animations on each element.  My suggestion would be to model this a bit more explicitly, rather than having that sequence modeled implicitly in the `setInterval` calls.  This might look like having `addMessages` put the new messages in state and have that be it, adding a callback to the VelocityComponent that tracks which messages have finished animating and stores that in state as `animatedMessageKeys`.  In the render code, all animated messages are rendered, along with the next message that hasn't been animated yet.  Rendering would cause the animation, and when it finishes that would lead to the next one being rendered, etc.  That would make the state changes a bit more explicit and you could remove the manual scheduling code to add new messages each 300ms.

Also, in general for you'll want to make sure that any callbacks scheduled via `setInterval` or `setTimeout` are cancelled when the component is unmounted.  Otherwise they can end up being scheduled to run, the user navigates away and the component unmounts, and then the callback runs but there's no more component left :)  Using [this mixin](https://github.com/mit-teaching-systems-lab/threeflows/blob/master/ui/src/helpers/set_interval_mixin.js) can help (or something similar).

kevinrobinson:
@david-rosales Nice work!  This is tricky stuff and your descriptions of what you tried are super helpful and your approach looks great.  I had a suggestion about how to simplify the staggering and chaining so there's less scheduling code, let me know if you want to chat it out, async and animations are always tricky.

david-rosales:
@kevinrobinson I changed the staggering to be queue-like in the sense that it will render the next message once the previous one finished animating. I used the `complete` option for velocity.

Can you review this again when you get a chance?

ui/src/message_popup/mobile_prototype/mobile_interface.jsx
```diff
@@ -44,38 +44,39 @@ export default React.createClass({
...
       exampleIndex: 0,
       dialog: null,
-      messageIndex: 1
+      messageIndex: 1,
     });
   },
 
   componentDidMount() {
     this.setInterval(this.updateTimer, ONE_SECOND);
+    this.onAnimationDone();
```
david-rosales: This was to get the initial message in the queue.

david-rosales:
@kevinrobinson I made the changes we discussed, can you review it?

kevinrobinson:
@david-rosales Approved, ship it!

PR #112, opened 2016-08-05
Removed sessionLength from Slider key

david-rosales:
The Slider was having fun breaking again. This time around, I think what was happening is that every time the slider changed, the key updated because `sessionLength` was included. This meant that the slider would be replaced every time it changed which led to console errors and no actual sliding, just clicking.

For reference, these were the two errors:
```
Uncaught TypeError: Cannot read property 'getBoundingClientRect' of undefined
```
```
Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the undefined component.
```


david-rosales:
@kevinrobinson Another temporary slider fix, can you review it?

kevinrobinson:
@david-rosales Yeah I just noticed this too :)  Nice fix, thanks for cleaning up by bug! 👍

PR #114, opened 2016-08-08
Created GET endpoint for message popup questions

david-rosales:
I set up the endpoint for querying the database for questions.

My thought for how the questions will be formatted is a parameter like the following in the questions section of the database row:
```
{currentQuestions: [ ... ], archivedQuestions: [ ... ]}
```

I believe to access this we would use `Api.questionsQuery().end(callback)` and `callback` would look like the following: 
```
function(err, response) {
  //... some checks here to make sure it actually exists ...
  const allQuestions = JSON.parse(response.text).rows[0]; //only 1 item
  const {currentQuestions, archivedQuestions} = allQuestions;
  //... setting the questions somewhere in state ...
}
```

For now, this pull request is just to see how it would work on the actual heroku server with actual access to the database.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -190,7 +190,8 @@ app.get('/server/questions', facultyAuth, function(request, response){
     const {rows} = result;
     console.log(rows);
     response.status(200);
-    return response.json({rows});
+    if(rows.length === 0) return response.json({});
+    return response.json({row: rows[0]});
   });
 });
 
--- a/server/index.js
+++ b/server/index.js
@@ -174,6 +174,26 @@ app.get('/server/evaluations', facultyAuth, function(request, response) {
   });
 });
 
+
+
+app.get('/server/questions', facultyAuth, function(request, response){
+  if (process.env.NODE_ENV === 'development') {
+    return response.json({});
+  }
+
+  queryDatabase('SELECT * FROM message_popup_questions ORDER BY timestamp DESC LIMIT 1', [], function(err, result){
+    if (err) {
+      console.log({ error: err });
+      return response.status(500);
+    }
+
+    const {rows} = result;
+    console.log(rows);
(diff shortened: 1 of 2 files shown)


david-rosales:
@kevinrobinson Can you review this to make sure I'm heading in the right direction?

server/index.js
```diff
@@ -174,6 +174,26 @@ app.get('/server/evaluations', facultyAuth, function(request, response) {
...
+  queryDatabase('SELECT * FROM message_popup_questions ORDER BY timestamp DESC LIMIT 1', [], function(err, result){
+    if (err) {
+      console.log({ error: err });
+      return response.status(500);
+    }
+
+    const {rows} = result;
+    console.log(rows);
+    response.status(200);
+    return response.json({rows});
```
kevinrobinson: Since we're grabbing the latest snapshot, we could also check that there's only one row and return that directly here.  Otherwise that step leaks to the client and it might be good to have the server code here encapsulate that.

ui/src/helpers/api.js
```diff
@@ -69,4 +69,11 @@ function joinEvaluationsWithEvidence([evaluationsResponse, evidenceResponse]) {
     const log = _.find(logs, log => log.id.toString() === evaluation.json.logId.toString());
     return {...evaluation, log};
   });
+}
+
+//Query for questions
+export function questionsQuery() {
+  return request
+    .get('/server/questions')
+    .set('Content-Type', 'application/json');
```
kevinrobinson: 👍

kevinrobinson:
@david-rosales This looks good!  If you'd like to get feedback and merge in small pieces as you go then go for it!  This is a good incremental improvement so totally cool to merge this in even though the whole feature isn't yet complete.

PR #115, opened 2016-08-09
Created POST endpoint for message popup questions

david-rosales:
I set up the post endpoint for the database table holding the questions.

It's a very small change but I just want to make sure it's not going to kill the database should I use it.

### DIFF
--- a/ui/src/helpers/api.js
+++ b/ui/src/helpers/api.js
@@ -78,10 +78,10 @@ export function questionsQuery() {
     .set('Content-Type', 'application/json');
 }
 
-export function saveQuestions(record) {
+export function saveQuestions(questions) {
   request
     .post('/server/questions')
     .set('Content-Type', 'application/json')
-    .send(record)
+    .send(questions)
     .end();
 }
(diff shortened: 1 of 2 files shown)


david-rosales:
@kevinrobinson Can you review this?

ui/src/helpers/api.js
```diff
@@ -76,4 +76,12 @@ export function questionsQuery() {
   return request
     .get('/server/questions')
     .set('Content-Type', 'application/json');
+}
+
+export function saveQuestions(record) {
```
kevinrobinson: Should `record` here be `questions`, based on how the server interprets the request body?

kevinrobinson:
@david-rosales One question on the request body format, but this looks good!

ui/src/helpers/api.js
david-rosales: Good point. I updated it but am wondering whether this has any actual effect on the underlying database code or is it just for readability?

kevinrobinson: Good question!  Just clarify, we could call it `variableNamesOnlyMatterToHumans`:)  But clarity in naming and documentation is usually particularly important when crossing system boundaries like here (eg., from the UI in the browser to the server code).

david-rosales: Sounds good, merging now.

PR #116, opened 2016-08-10
Remove duplicate call to renderQuestion

kevinrobinson:
Ran into what looks like a merge bug in adding audio responses.

### DIFF
--- a/ui/src/message_popup/popup_question.jsx
+++ b/ui/src/message_popup/popup_question.jsx
@@ -148,7 +148,6 @@ export default React.createClass({
     return (
       <div>
         <VelocityTransitionGroup leave={{animation: "slideUp"}} runOnMount={true}>
-          {!scaffolding.shouldShowSummary && this.renderQuestion()}
           {!isDoneRevising && scaffolding.shouldShowSummary && this.renderQuestion()}
         </VelocityTransitionGroup>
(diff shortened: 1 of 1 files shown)


kevinrobinson:
@david-rosales Can you review?  I think this extra line came in from a merge and wasn't doing any real harm but doesn't need to be there.

ui/src/message_popup/popup_question.jsx
```diff
@@ -148,7 +148,6 @@ export default React.createClass({
     return (
       <div>
         <VelocityTransitionGroup leave={{animation: "slideUp"}} runOnMount={true}>
-          {!scaffolding.shouldShowSummary && this.renderQuestion()}
           {!isDoneRevising && scaffolding.shouldShowSummary && this.renderQuestion()}
```
david-rosales: @kevinrobinson I actually think this line was needed.
Line 151 renders the question if the summary does not get shown and line 152 renders the question if the summary should be shown but the question is still in progress. 

If you toggle off showing summary, you get what happens below.

line 151 removed vs line 151 included.
<div style="display: flex; flex-direction: row">
screen shot 2016-08-10 at 10 39 41 am
screen shot 2016-08-10 at 10 40 02 am

</div>

kevinrobinson: ah thanks, great review! 👍 I'll drop this then and am looking at refactoring some of this for audio responses and will clarify separately.

PR #117, opened 2016-08-10
Read AWS credentials for audio as single env variable

kevinrobinson:
This is a revision to https://github.com/mit-teaching-systems-lab/threeflows/pull/100, to keep the format in local dev and production the same (just values are different).


kevinrobinson:
selfie, updating this in Heroku now cc @david-rosales

PR #118, opened 2016-08-10
Setting up Question Authoring with the database

david-rosales:
I started implementing the usage of the database endpoints in the Question Authoring. Merging this will effectively hide all currently listed questions as it should take its data strictly from the database when available. 

One issue I ran into and have made a temporary solution for is that the `QuestionPage` component that gets rendered into the `EditQuestionPage` component must be rendered only when the `originalQuestion` is ready to be passed in. Otherwise, the `getInitialState` of `QuestionPage` will treat the question, even if coming from `EditQuestionPage`, as a new page because the `originalQuestion` is still not set and seen as `undefined`. To fix this, I just rendered the `QuestionPage` once `originalQuestion` was no longer `undefined`. Interestingly, when I did this, `Api.questionsQuery().end(this.onQuestionsReceived);` was no longer running even though `componentDidMount` ran perfectly fine. For now, since I mainly want to test the database endpoints I just passed in a prop will `allQuestions` which gets calculated anyways in `EditQuestionPage`. Later, I think we should look into why this was happening. The `questionsQuery` calls were still left in `QuestionPage` because it was still needed as a new question would not be able to pass in the `allQuestions` prop.

As I mentioned, I mainly want to test the database endpoints first to make sure they are working and that questions could be created, editing, and deleted (archived).


david-rosales:
@kevinrobinson Can you review this when you get a chance?

ui/src/message_popup/author/question_page.jsx
```diff
@@ -25,21 +27,30 @@ export default React.createClass({
 
   propTypes: {
     originalQuestion: React.PropTypes.object,
+    allQuestions: React.PropTypes.object
```
kevinrobinson: Could we remove this altogether from `QuestionPage`?  It looks like the only place it's used is to figure out the next `questionId` when creating a new one.  I think it will simplify to leave the actual saving as the responsibility of a container outside of `question_page`, and have it handle making ids and communicating with the database when it's time to save, edit or delete.

So concretely, this would mean refactoring methods here like `onSaveEdit` so that they all ultimately call something like `onQuestionsChanged`.  This will reduce some of the duplication between then now.  We can then factor the `onQuestionChanged` logic out of this component so that `QuestionPage` stays focused on only dealing with a single question.  The `onQuestionsChanged` logic can then be pulled out into the top-level `new_question_page` or `edit_question_page` components.

Let me know if I'm explaining this well or if it's easier to talk out in person!

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -29,12 +30,18 @@ export default React.createClass({
 
   getInitialState(){
     return ({
-      questions: allQuestions,
+      allDatabaseQuestions: [],
+      currentQuestions: [],
+      archivedQuestions: [],
```
kevinrobinson: Choose a single representation for this state, rather than having multiple.  It's up to you whatever shape of data works best (eg., a single list, a map with two lists, two lists) but avoid storing two different projections of the same data in component `state`.

I think it probably makes the most conceptual sense to treat this as a single entity (either a single list or a map with two lists), since it will be loaded as one thing from the server.

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -59,18 +66,37 @@ export default React.createClass({
     return questionString.split(text).length;
   },
   
+  onQuestionsReceived(err, response){
+    const allDatabaseQuestions = JSON.parse(response.text).row;
```
kevinrobinson: This is related to what we were talking about in https://github.com/mit-teaching-systems-lab/threeflows/pull/114/files#diff-0861d6d6b50d7d695344bf2d86d6e5e6R194, although perhaps I missed it earlier.  Seeing `row` here feels like it's a detail of how we're storing things in the database, and doesn't help communicate what this data actually is.

Based on the code here, it looks like the endpoint will return `allDatabaseQuestions` which is an object like `{ currentQuestions: [], archivedQuestions: [] }`.  So you could remove the `row` wrapped in the server and in the parsing here, or you could change the `row` wrapper to `allDatabaseQuestions` to communicate what it is.

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -59,18 +66,37 @@ export default React.createClass({
...
+    if(allDatabaseQuestions === undefined) {
```
kevinrobinson: I don't quite follow why the branching here is needed.  Maybe simplifying the shape of `getInitialState` would help?  We should be able to get the flow working where: `component mounts -> requests data -> gets data -> sets it in state` in one pass.

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -59,18 +66,37 @@ export default React.createClass({
...
     const value = event.target.value.toLowerCase().trim();
     if(value === ''){
-      this.setState({questions: allQuestions});
+      this.setState({currentQuestions: this.state.allDatabaseQuestions.currentQuestions});
       return;
     }
-    const questions = allQuestions.filter(questionOriginal => this.searchQuestionForText(value, questionOriginal)).sort(questionOriginal => this.countTextOccurrences(value, questionOriginal));
-    this.setState({questions});
+    const questions = this.state.allDatabaseQuestions.currentQuestions.filter(questionOriginal => this.searchQuestionForText(value, questionOriginal)).sort(questionOriginal => this.countTextOccurrences(value, questionOriginal));
+    this.setState({currentQuestions: questions});
```
kevinrobinson: Previously, there were two different concepts: `allQuestions` and `questions`.  Now that we're updating this and the set of questions is fetched from the server and can change we'll want to keep these as distinct things.

I should have noticed this previously, but in React components in general, filtering and other computations should be in the `render` path whenever possible.  In this case we can compute the list to show from `currentQuestions` and the search text.  If this method stores the search text in component state, then we can describe this minimally with only two concepts, and the filtered list can be computed from them.

ui/src/message_popup/transformations.jsx
```diff
@@ -20,8 +20,9 @@ export function withIndicator(question) {
   };
 }
 
-export function questionsForIndicator(indicatorId) {
-  const withIndicators = _.compact(allQuestions.map((question) => {
+export function questionsForIndicator(indicatorId, questions) {
+  const questionsList = (questions === undefined) ? allQuestions : questions;
```
kevinrobinson: I'd make this more strict, so that you have to pass in `questions`  That may involve updating a few call sites but makes explicit that this module can't be expected to get `allQuestions` on its own anymore and callers have to handle that.

ui/src/message_popup/author/edit_question_page.jsx
```diff
@@ -14,12 +16,45 @@ export default React.createClass({
...
+      }
+    });
+  },
+
+  componentDidMount(){
+    Api.questionsQuery().end(this.onQuestionsReceived);
+  },
+
+  onQuestionsReceived(err, response){
+    const allDatabaseQuestions = JSON.parse(response.text).row;
```
kevinrobinson: I don't quite follow what's happening here (same comments as in `questions_page.jsx`).

ui/src/message_popup/author/edit_question_page.jsx
```diff
@@ -14,12 +16,45 @@ export default React.createClass({
...
+    if(allDatabaseQuestions === undefined) {
+      this.setState({
+        allDatabaseQuestions: {
+          currentQuestions: allQuestions, 
+          archivedQuestions: allArchivedQuestions
+        }
+      });
+      return;
+    }
+    this.setState({ allDatabaseQuestions: allDatabaseQuestions });
```
kevinrobinson: Use destructuring here if you have the same key and variable names.

kevinrobinson:
@david-rosales Nice explanation and comments, that really helped me follow what you were thinking about 👍 

I'd suggest focusing on https://github.com/mit-teaching-systems-lab/threeflows/pull/118/files#r74449837 first, and then after that seeing if you can simplify the way the questions state is stored and updated in components, and let me know if you want to talk it out at all.

ui/src/message_popup/author/questions_page.jsx
david-rosales: The reasoning behind separating it out in the state like above is because `allDatabaseQuestions` would remain constant throughout the component's lifecycle, (only changing when`onQuestionsChanged` would be called) but the `currentQuestions` and `archivedQuestions` would add and remove questions depending on search bar input. The full list of questions however, found in `allDatabaseQuestions`, would be needed to restore the `currentQuestions` and `archivedQuestions` lists after the search bar changes again or is cleared.

I now see though from your comment on getting these search bar lists in the render method that I can definitely simplify this.

ui/src/message_popup/author/questions_page.jsx
david-rosales: It was initially there to set the question lists if the database was not connected. I do see now that it could be set in `getInitialState` and overwritten when getting the data from the database.

ui/src/message_popup/author/questions_page.jsx
kevinrobinson: Ah yeah that totally makes sense, and yep the simplification comes from deferring computation of filtering the list to render, hopefully that helps simplify :)

ui/src/message_popup/author/questions_page.jsx
kevinrobinson: Ah got it, thanks for explaining!  Yeah the other option is to have the top-level page handle the fetching and loading.  And if it hasn't gotten back the list of questions from the server yet it can show 'Loading' or something like that and not render the `<QuestionPage />` yet.  And then when the server request comes back and it sets it in the component's state, `render` can check that and only then render the `QuestionPage`.  That way this component can require it always be passed as a prop.

kevinrobinson: That's a common pattern for loading data from the server - have one "container" component handle fetching data, tracking that state, and then branching between a loading state a "pure" UI component that always requires all the props and handles showing the data to the user.

david-rosales:
@kevinrobinson Can you review this again when you get a chance?

ui/src/message_popup/transformations.jsx
david-rosales: Since this would change how other non-authoring files gather the questions, I decided I would handle it in a separate pull request where we start using the database questions in the actual message popup experience.

ui/src/message_popup/author/edit_question_page.jsx
```diff
@@ -2,7 +2,9 @@
 import React from 'react';
 import _ from 'lodash';
 
+import * as Api from '../../helpers/api.js';
 import {allQuestions} from '../questions.js';
+import {allArchivedQuestions} from './archived_questions.js';
 import {withStudents, withIndicator} from '../transformations.jsx';
 
 import QuestionPage from './question_page.jsx';
```
kevinrobinson: Please update the component name and filename so they match, otherwise grepping and searching gets tricky :)

ui/src/message_popup/author/edit_question_page.jsx
```diff
@@ -14,12 +16,49 @@ export default React.createClass({
...
+    const allQuestionsFlat =  this.state.allDatabaseQuestions.currentQuestions.concat(this.state.allDatabaseQuestions.archivedQuestions);
+    const originalQuestion = _.find(withStudents(allQuestionsFlat).map(question => withIndicator(question)), question => question.id.toString() === this.props.questionId);
     return (
-      <QuestionPage 
-        originalQuestion={question}
-        />
-      );
+      <div>
+        {!loaded &&
+          <QuestionPage
```
kevinrobinson: I don't understand this branch, can you explain?  I was expecting it to just show "loading..." or something until the data comes back from the server.

ui/src/message_popup/author/new_question_page.jsx
```diff
@@ -0,0 +1,46 @@
...
+import QuestionPage from './question_page.jsx';
+import {allArchivedQuestions} from './archived_questions.js';
+import {allQuestions} from '../questions.js';
+import * as Api from '../../helpers/api.js';
+
+
+export default React.createClass({
+  displayName: 'NewQuestionPage',
+
+  getInitialState(){
```
kevinrobinson: After this PR, let's talk out or pair to factor out the "loading questions from the server" bit of this component, `new_question_page` and `questions_page`.  We can make what's called a "higher-order component" that does this and then passes along the questions to child components.  That will help simplify and split out the loading code into one component.

ui/src/message_popup/author/new_question_page.jsx
```diff
@@ -0,0 +1,46 @@
...
+    return ({
+      loaded: false,
+      allDatabaseQuestions: {
```
kevinrobinson: I'm not following why `allQuestions` and `allArchivedQuestions` are here in `getInitialState`, can this be removed now that this loads them from the server?

ui/src/message_popup/author/edit_question_page.jsx
```diff
@@ -14,12 +16,49 @@ export default React.createClass({
     questionId: React.PropTypes.string.isRequired
   },
 
+  getInitialState(){
```
kevinrobinson: Same comments as elsewhere about some of the loading code.

ui/src/message_popup/author/new_question_page.jsx
```diff
@@ -0,0 +1,46 @@
...
+    });
+  },
+
+  componentDidMount(){
+    Api.questionsQuery().end(this.onQuestionsReceived);
+  },
+
+  onQuestionsReceived(err, response){
+    const questions = JSON.parse(response.text).questions;
+    if(questions !== undefined){
```
kevinrobinson: Is this to handle if there's nothing in the database?  If so you could also check for `err` here.  In either case, setting `loaded: true` will then cause the rendering code to read in the undefined value.

Another approach could be encoding the state of the request (rather than just the boolean `loaded`).  That could look like `requestState` being 'idle' | 'loading' | 'error'

ui/src/message_popup/author/question_page.jsx
```diff
@@ -25,6 +25,8 @@ export default React.createClass({
 
   propTypes: {
     originalQuestion: React.PropTypes.object,
+    allQuestions: React.PropTypes.object.isRequired,
+    loaded: React.PropTypes.bool.isRequired
```
kevinrobinson: Why is `loaded` necessary here?  And why all the questions?

ui/src/message_popup/author/question_page.jsx
```diff
@@ -104,8 +106,11 @@ export default React.createClass({
     console.log("Saving the following new question:");
     const studentIds = this.state.students.map(student => student.id);
     const text = this.state.questionText;
+    const {originalQuestion} = this.props;
     const {goodExamples, badExamples} = this.getExamples();
-    const id = _.maxBy(allQuestions, question => question.id).id + 1;
```
kevinrobinson: Let's talk out in person how this component could require a prop like `onQuestionEdited`.  If we did that, I think we could remove all this code and the need for this component (which is about editing a single question) to have to know about all the questions in the database.  The component above could handle that.

Similar for `onCreate` and `onDelete`.

ui/src/message_popup/author/question_page.jsx
```diff
@@ -150,62 +166,75 @@ export default React.createClass({
...
+    }
+    Api.saveQuestions({currentQuestions, archivedQuestions});
+    this.setState({currentQuestions, archivedQuestions});
     this.onReturnToQuestions();
   },
 
   render(){
+    const {loaded, originalQuestion} = this.props;
+    const {questionText, students, goodExamplesText, badExamplesText, indicator, deleteConfirmationOpen} = this.state;
+    const title = originalQuestion !== undefined ? `Editing Question #${originalQuestion.id}` : loaded ? 'New Question' : '';
```
kevinrobinson: Can we remove `loaded` altogether, and have the parent component not render `QuestionPage` until it has loaded the question?

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -29,12 +31,21 @@ export default React.createClass({
 
   getInitialState(){
     return ({
-      questions: allQuestions,
```
kevinrobinson: Same questions as elsewhere about loading the questions data.

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -59,26 +70,54 @@ export default React.createClass({
...
+  getSearchResults(){
+    const {searchText} = this.state;
+    if(searchText === '') return this.state.allDatabaseQuestions;
+    const currentQuestions = this.state.allDatabaseQuestions.currentQuestions
+      .filter(question => this.searchQuestionForText(searchText, question))
+      .sort(question => this.countTextOccurrences(searchText, question));
+    const archivedQuestions = this.state.allDatabaseQuestions.archivedQuestions
+      .filter(question => this.searchQuestionForText(searchText, question))
+      .sort(question => this.countTextOccurrences(searchText, question));
+    return {currentQuestions, archivedQuestions};
```
kevinrobinson: This looks good!  I'd suggest factoring the calls to filter/sort out into a method, since it's the same for both sets of questions.

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -59,26 +70,54 @@ export default React.createClass({
...
+
   onNewQuestion(){
     Routes.navigate(Routes.messagePopupAuthorQuestionsNewPath());
   },
 
+  onQuestionRestore(){
+    const {selectedArchivedQuestion} = this.state;
+    if(selectedArchivedQuestion !== null){
+      const allDatabaseQuestions = _.clone(this.state.allDatabaseQuestions);
+      _.remove(allDatabaseQuestions.archivedQuestions, question => question.id === selectedArchivedQuestion.id);
```
kevinrobinson: I'm not sure this matters in terms of performance or memory usage, and this may just be a style preference, but   I'd suggest using transformations over mutations whenever you can (here that would be `.filter`and `.concat` instead of `_.clone`, `_.remove`, `.push`.

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -59,26 +70,54 @@ export default React.createClass({
...
+      allDatabaseQuestions.currentQuestions.push(selectedArchivedQuestion);
+      Api.saveQuestions(allDatabaseQuestions);
```
kevinrobinson: This is a fire-and-forget call to the server, and we should check the return value here but let's do that in another pass after this pull request.

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -59,26 +70,54 @@ export default React.createClass({
...
+
   onSearchBarChange(event){
     const value = event.target.value.toLowerCase().trim();
-    if(value === ''){
-      this.setState({questions: allQuestions});
-      return;
-    }
-    const questions = allQuestions.filter(questionOriginal => this.searchQuestionForText(value, questionOriginal)).sort(questionOriginal => this.countTextOccurrences(value, questionOriginal));
-    this.setState({questions});
+    this.setState({searchText: value});
```
kevinrobinson: 👍

kevinrobinson:
@david-rosales This is great progress, and good simplification in this pass 👍 

For this PR, let's finish up:
[ ] handling the loading state while requesting data from the server
[ ] moving the logic to update/create/delete questions to the components above `QuestionEdit`

And if you want we can pair on splitting out the loading state code into its own component now, or we could do that in a separate pass after, whichever you prefer.

ui/src/message_popup/transformations.jsx
kevinrobinson: 👍

ui/src/message_popup/author/edit_question_page.jsx
david-rosales: Which component name are you referring to?

ui/src/message_popup/author/edit_question_page.jsx
david-rosales: The reasoning behind the `loaded` variable was to make the transition between the pages more fluid and not show any white screens (even though they may quickly disappear) while transitioning and waiting for the questions from the database.

It gets passed down so that `QuestionPage` can render its app bar and then wait for the questions from the database. Once the questions are received, it will load the rest of the components which get animated in so the transition looks smooth.

ui/src/message_popup/author/new_question_page.jsx
david-rosales: This was mainly for UI testing purposes until I know for sure that the database setup is working.

ui/src/message_popup/author/new_question_page.jsx
david-rosales: Yes, I initially tried checking for `err` but I found that it kept being null even the database never connected. Since I'm not too familiar with how it should have worked, I decided to do a workaround. 

With this, it passes down `allDatabaseQuestions` as a prop. Since the `onQuestionsReceived` only updates the `allDatabaseQuestions` prop, if it is initially set as something else, setting loaded to true should not render anything undefined but what was set in `getInitialState` until overridden by a successful database query.

ui/src/message_popup/author/question_page.jsx
david-rosales: I think adding the higher level component above this which would also handle the question loading instead of `NewQuestionPage` and `EditQuestionPage` would work.

ui/src/message_popup/author/edit_question_page.jsx
kevinrobinson: ah sorry I read this wrong, my mistake!

ui/src/message_popup/author/edit_question_page.jsx
kevinrobinson: Ah got it, thanks!  This makes sense then, or if it's just the `AppBar` that should render even during loading, you could pull that up to the top level component here.  The benefit being that loading states is removed from `QuestionPage` altogether, but the way you have it here looks good too.

ui/src/message_popup/author/new_question_page.jsx
kevinrobinson: 👍

ui/src/message_popup/author/new_question_page.jsx
kevinrobinson: Yeah having the questions in `getInitialState` makes sense for initial UI testing, but that won't be like that going forward.

For err, the `API` methods are ultimately making HTTP requests using a library called superagent, and it follows the node style for passing errors back, which you can see more about here: https://visionmedia.github.io/superagent/#error-handling

I think the root issue may be that we should update the endpoint you added in https://github.com/mit-teaching-systems-lab/threeflows/pull/114 so that it returns `{ questions: [] }` to match the changes in this PR.

ui/src/message_popup/author/question_page.jsx
kevinrobinson: 👍 Let's do it!  And to clarify a bit I think there's two separate issues: one is the "how do I load questions from the server and then pass them to a child component once they're loaded" and the other is "when an update/create/delete is made to the set of questions, how do I send those back to the server and update the list of questions in the browser."

ui/src/message_popup/author/new_question_page.jsx
david-rosales: Ok, I now see why I `err` was `null`. If it is in development mode it returned `response.json({})` which caused `JSON.parse(response.text).questions` to be `undefined`. It shouldn't happen in the actual build so I'll change it to search for `err` instead of checking for `undefined`. Also, if in development mode (mainly because a failed get query to the database crashes my localhost, I'm not too sure if this is universal) I think we should just return an empty questions object structured to be like the actual one:
```
response.json({questions: {currentQuestions: [], archivedQuestion: []}})
```

kevinrobinson: Yeah, that all sounds great! 👍

kevinrobinson: Also we could make this more resilient to errors, which are good improvements to make in general, but I think biasing towards "let's just get this working first" is a good mindset at this point.

ui/src/message_popup/author/edit_question_page.jsx
```diff
@@ -1,8 +1,8 @@
-/* @flow weak */
```
david-rosales: Flow was having issues with the props possibly being undefined. I couldn't fix it with setting the props as required because then React was having issues because I passed them in as a cloned element using `React.cloneElement`. I also tried using to prevent the props from possibly being undefined `getDefaultProps` but flow wouldn't recognize that either. For now, I disabled flow on this file and two others with the same issue.

ui/src/message_popup/author/new_question_page.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint no-console: "off" */
```
david-rosales: Didn't exactly remove flow here but just didn't add it for the same reason I noted above.

ui/src/message_popup/author/questions_page.jsx
```diff
@@ -1,4 +1,3 @@
-/* @flow weak */
```
david-rosales: Removed for the reason stated above.

ui/src/message_popup/author/question_page.jsx
```diff
@@ -99,46 +102,34 @@ export default React.createClass({
...
       nonExamples: badExamples,
       indicatorId: this.state.indicator.id,
-      indicator: this.state.indicator,
+      indicator: this.state.indicator
     };
-    console.log(question);
-    console.log("Also deleting and archiving the following question:");
-    console.log(this.props.originalQuestion);
-    console.log("");
+    if(this.props.onEditQuestion === undefined) return;
```
david-rosales: I tried returning at the beginning of the method if it was not defined but flow wouldn't recognize it unless it was right before when the method was called.

david-rosales:
@kevinrobinson Can you review this when you have a chance?

ui/src/message_popup/author/edit_question_page.jsx
kevinrobinson: Cool, sounds good to just work around for now and I can help take a look with you later if you like.

ui/src/message_popup/author/new_question_page.jsx
kevinrobinson: 👍  Cool, let's tackle separately.  Can you remove the eslint suppression now though?

david-rosales: Because the function to create a new question was moved here and it logs what its doing to the console, there was a need for the eslint suppression. However, I'm pretty sure by now from what it has already logged to the console in the past that the functions are working properly so I can remove the logging entirely.

kevinrobinson: Ah yeah my bad I forgot about that, totally makes sense to leave for now until we finish this off.

ui/src/app.jsx
```diff
@@ -113,15 +114,27 @@ export default React.createClass({
   },
 
   messagePopupAuthorQuestions(query = {}) {
-    return <MessagePopup.QuestionsPage />;
+    return (
+      <QuestionsLoader>
```
kevinrobinson: I'd prefer the more explicit form where you pass a render function in as a prop for something like this (since this form hides how the props get passed into the other components), but this works too.

ui/src/message_popup/author/edit_question_page.jsx
```diff
@@ -11,15 +11,67 @@ export default React.createClass({
...
   propTypes: {
-    questionId: React.PropTypes.string.isRequired
+    questionId: React.PropTypes.string.isRequired,
+    loaded: React.PropTypes.bool,
+    allQuestions: React.PropTypes.object,
+    onReloadQuestions: React.PropTypes.func
+  },
+
+  componentDidMount(){
+    this.props.onReloadQuestions();
```
kevinrobinson: Does making the initial request make more sense in the loader (so it would always request on mounting)?

ui/src/app.jsx
kevinrobinson: Oh also let's remove any logic at all from the `app` page so that it only has responsibility for deciding what page to render.  Since all the routes have to go through the `app` page it will continue to grow for new pages we add, so keeping everything else out can help make it more manageable.

So these can move into `NewQuestionPage` etc.

ui/src/message_popup/author/new_question_page.jsx
```diff
@@ -0,0 +1,50 @@
...
+    return largestID + 1;
+  },
+
+  onCreateQuestion(newQuestion){
+    const {currentQuestions, archivedQuestions} = this.props.allQuestions;
+    const question = {id: this.getNewID(), ...newQuestion};
+    const newCurrentQuestions = currentQuestions.concat(question);
+    console.log('Creating new question...');
+    console.log(question);
+    Api.saveQuestions({currentQuestions: newCurrentQuestions, archivedQuestions});
```
kevinrobinson: Ahh, I think I see more why you were talking about mixing the "update/edit/create" capability with the loader capability, since this needs a way to update the state of questions as well as making the server call.

Let's try this approach: if we move the loader component into each page (out of `app`), and add another prop to `QuestionLoader` for updating the questions (eg., `onQuestionsUpdated`), we can the move the server call into the loader as I think you originally suggested.

ui/src/message_popup/author/edit_question_page.jsx
david-rosales: I originally had that but `onReloadQuestions` would not be called after rerouting. The most important one being not reloading the questions after editing a question and returning to `QuestionsPage`. I think it's because React swaps out one `QuestionsLoader` with another identical one and it has no indication that the `QuestionsLoader` component actually changed and so it is 'already mounted'. Maybe adding keys to the `QuestionsLoader`s would fix this but I felt the most surefire way of making sure that the questions are reloading when necessary is doing so in the components that need them.

ui/src/message_popup/author/new_question_page.jsx
david-rosales: I'm not too sure if I follow here. Would `onQuestionsUpdated` be replacing `onCreateQuestion`, `onEditQuestion`, and `onDeleteQuestion`? `EditQuestionPage` also needs `allQuestions` to grab the question being edited from an id, would this change?

Also, how would we handle `QuestionsPage` if we moved `QuestionsLoader` out of `app` since it needs `allQuestions` from the loader?

One thing that does make sense to me is passing a function as a prop into `QuestionsLoader` that when called updates `loaded` and `allQuestions` in `EditQuestionPage`, `NewQuestionPage`, and `QuestionsPage` but I don't think that's what you meant.

ui/src/message_popup/author/edit_question_page.jsx
kevinrobinson: Ah got it, great diagnosis and explanation.  Yeah I don't think they'd need to reload on that navigation if we update their values directly when changing the questions.  Let's talk it out in person.

ui/src/message_popup/author/new_question_page.jsx
kevinrobinson: Let's talk it out :)

ui/src/message_popup/author/edit_question_page.jsx
```diff
@@ -11,15 +11,96 @@ export default React.createClass({
   displayName: 'EditQuestionPage',
 
   propTypes: {
-    questionId: React.PropTypes.string.isRequired
+    questionId: React.PropTypes.string.isRequired,
+    loaded: React.PropTypes.bool,
+    allQuestions: React.PropTypes.object
+  },
+
+  getDefaultProps(){
```
david-rosales: I used this format of `getDefaultProps` for the files that load the questions from the database so that we can write tests for them with props instead of just having everything in the state.

david-rosales:
@kevinrobinson I removed `QuestionsLoader` and moved the question loading back to the components. Doing so was still helpful because it helped organize the code and now there are only three separate calls to the database in three files but it was much easier to implement this time around than the first time.

Can you review this so we can get it merged and tested?

ui/src/message_popup/author/edit_question_page.jsx
kevinrobinson: This is a good idea based on our conversation yesterday, but I think for this scenario we'll be better served by mocking the calls to the server.  As an example, in test we could let this component make calls through the Api module and use [sinon](http://sinonjs.org/docs/) to spy on that.  Or we can allow passing in an instance of `api` as a prop, and then pass in one that has all spies in test.

This seems like a better approach to me for this particular component, because the essence of what it is doing is making those calls to the API, and so if we have a test that passes in props then we don't test any of that code.

So I'd suggest removing these props, and then if this seems in good shape we can also merge this and add in additional tests in a separate pass (since this feature isn't in production use yet).

ui/src/message_popup/author/edit_question_page.jsx
```diff
@@ -11,15 +11,96 @@ export default React.createClass({
...
+    const allQuestions = JSON.parse(response.text).questions;
+    this.setState({ loaded: true, allQuestions });
+  },
+
+  onEditQuestion(newQuestion, originalQuestion){
+    const {currentQuestions, archivedQuestions} = this.state.allQuestions;
+    const question = {id: this.getNewID(), ...newQuestion};
+    const newCurrentQuestions = currentQuestions.filter(question => question.id !== originalQuestion.id).concat(question);
+    const newArchivedQuestions = archivedQuestions.concat(originalQuestion);
+    console.log('Replacing...');
```
kevinrobinson: This method and `onDeleteQuestion` should also update the questions in `state`, right?

ui/src/message_popup/author/new_question_page.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+export default React.createClass({
+  displayName: 'NewQuestionPage',
+
+  propTypes: {
+    loaded: React.PropTypes.bool,
+    allQuestions: React.PropTypes.object,
+  },
+
+  getDefaultProps(){
```
kevinrobinson: Same suggestion here about removing the props, and testing in a way that will exercise the queries to the server.

kevinrobinson:
@david-rosales This looks good, awesome work going through the iterations to keep improving this!  Let me know if you want to talk more about the suggesting for testing too.

ui/src/message_popup/author/edit_question_page.jsx
david-rosales: I thought about doing so but it will be getting redirected to `QuestionsPage` immediately after so I felt that updating the state was unnecessary. `QuestionsPage` should query the database and get the updated questions list.

kevinrobinson: Ah got it, that's great, and we could optimize to remove the navigation and re-fetching separately.

david-rosales:
@kevinrobinson Merge time?

kevinrobinson:
@david-rosales 
image

PR #119, opened 2016-08-11
Factoring out audio recorder from React component shim, clarifying when microphone is needed

kevinrobinson:
This is a refactor to accomplish a small product improvement.  Previously, `AudioCapture` could be used as a controlled component to record audio blobs.  But the way it was implemented, it would add in the AuO component immediately, which prompted the user to use the audio and indicated that audio was being captured even when it wasn't actually yet.  This is because AuO captures the stream for visualization separately from recording.

We should investigate replacing this altogether, but this at least defers the "audio is recording" browser UI indication until the user clicks "record."

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   },
   "dependencies": {
     "aws-sdk": "^2.4.8",
+    "babel-plugin-transform-class-properties": "^6.11.5",
     "basic-auth": "^1.0.4",
     "body-parser": "^1.15.1",
     "date-fns": "^1.3.0",
--- a/ui/.babelrc
+++ b/ui/.babelrc
@@ -4,6 +4,7 @@
     "development": {
       "plugins": [
         ["transform-flow-strip-types"],
+        ["transform-class-properties"],
         ["react-transform", {
           "transforms": [{
             "transform": "livereactload/babel-transform",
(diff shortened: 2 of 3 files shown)


kevinrobinson:
@david-rosales Can you review when you have a chance?

david-rosales:
@kevinrobinson Looks good but one issue I had running it is that I ran into the error below: 

```
SyntaxError: /Users/anekin/Desktop/Repositories/TSL/threeflows/ui/src/components/audio_capture.jsx: 
Missing class properties transform. (line:13)
```

It would crash the `npm run watch` process on startup. I fixed this by adding: 
```
["transform-class-properties"]
```
to the `plugins` section of the `.babelrc`.

I also had to install the following: 
```
threeflows $ npm install --save babel-plugin-transform-class-properties
```

After doing those two things I was able to run the code and actually get past the actual build and the audio is recording icon did not show up until audio recording began as intended.

kevinrobinson:
@david-rosales Thanks for the awesomely helpful review and catching that and helping fix it up! 😄    I'm surprised this didn't fail the Travis build when I submitted the PR but didn't look into that further at the moment (but it would be good to someday if something similar comes up again).

PR #120, opened 2016-08-15
Add Mocha and Enzyme test setup, add initial tests for HintCard

kevinrobinson:
This adds Mocha for JavaScript tests, using Chai for assertions and Enzyme for some React-related utilities.  I'm not sure that there's a whole lot of value in Enzyme at the moment, but we can give it a try in the first few tests and see.  There's a separate branch experimenting with Jest but that ran into more snags during the initial setup so I think this is better for getting started right now.  It'd be good to take another look at this after we start with a few tests and it'd be a good experiment to look at adopting Create React App which now includes Jest too.  But that's a project for another day.

This isn't fully complete, and some important features that I didn't play with yet are:
- integration-style tests that render more of the component tree
- mocking server calls
- testing rendering inside VelocityTransitionGroup
- debugging in web inspector
- warning only if prop is missing or the wrong type


kevinrobinson:
@david-rosales Want to check this out and then pair on making some initial tests at some point this week?  Maybe some pieces of the authoring UI would be a good place to start.

ui/test/message_popup/hint_card_test.jsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
```
david-rosales: Is this file a duplicate of `ui/test/hint_card_test.jsx`?

ui/test/message_popup/hint_card_test.jsx
```diff
@@ -0,0 +1,75 @@
...
+import {mount, shallow} from 'enzyme';
+import {expect} from 'chai';
+import sinon from 'sinon';
+
+import RaisedButton from 'material-ui/RaisedButton';
+import HintCard from '../src/message_popup/hint_card.jsx';
```
david-rosales: I think this import might fail because it would search for `src/message_popup/hint_card.jsx` inside of `ui/test`.

 Would this work instead?
```
../../src/message_popup/hint_card.jsx
```

ui/test/hello_react_test.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
david-rosales: Should this file be saved as `hello_react_test.jsx` instead of `hello_react_test.js`?

david-rosales:
@kevinrobinson There's a few small things that travis has picked up/might pick up. Other than that, it looks good. Also, lets definitely find some time this week to pair on this.

ui/test/message_popup/hint_card_test.jsx
kevinrobinson: Yep, good catch!  Fixin'

ui/test/hello_react_test.js
kevinrobinson: Yep, updated!  This also led to adding `mocha.opts` which specifies what glob to use when looking for test files.

ui/test/message_popup/hint_card_test.jsx
kevinrobinson: Yeah sorry, moving these files was a last-minute refactoring and I didn't check it before.  Thanks for the helpful review!

PR #121, opened 2016-08-15
Refactor Question component into different renderer components

kevinrobinson:
This refactors the `PopupQuestion` component, splitting out separate components:
- `ScenarioRenderer` for showing the representation of the scenario
- `PromptsRenderer` for showing prompts and cues
- `ResponseRenderer` for allowing the user to respond with some approximation

The intent here is to isolate code that handles details about a particular representation of the scenario and response approximations so that we can add additional formats here (eg., video representation, audio response approximations).

There are no functional changes here, other than a minor change to `FeedbackCard` that came up in the process, where the user can't choose to "Revise" their response until they actually type something different.


kevinrobinson:
@david-rosales Can you review this refactor?  The idea is that this is the first step to then add in branching for these different renderers that will show video for some question types, and audio responses for some others.

A separate question (which I didn't think about here yet) is whether this structure would work well with the texting-based UI, and whether we could essentially remove `MobileInterface` and express the texting-based UI as a different `ScenarioRenderer` and `ResponseRenderer`.  I'd be curious to hear what you think about that separate from just merging this PR but it seems like a good time to bring it up. :)

ui/src/message_popup/popup_question.jsx
```diff
@@ -49,221 +47,88 @@ export default React.createClass({
...
-    this.onDonePressed();
+  onResponseSubmitted(response) {
+    this.setState({response});
   },
 
-  onRequestClose() {
-    //This empty function is meant to cancel out the closing of the toast without having to use
-    //an extremely large autoHideDuration
+  onDone() {
+    this.props.onDone(this.state.response.elapsedMs * 1000);
```
david-rosales: Is the `*1000` necessary? Isn't it already in milliseconds?

ui/src/message_popup/popup_question.jsx
```diff
@@ -49,221 +47,88 @@ export default React.createClass({
...
+    const {question, scaffolding, isLastQuestion} = this.props;
+
     return (
       <SummaryCard 
-        onDone={this.onDonePressed} 
-        question={this.props.question} 
-        response={this.props.scaffolding.helpType==='feedback' ? this.state.finalResponseText : this.state.initialResponseText} 
-        elapsedSeconds={Math.round(this.state.elapsedMs / 1000)} 
-        buttonLabel={this.props.isLastQuestion ? "Finish" : "Next Question"}/>
+        onDone={this.onDone}
```
david-rosales: `onDone` is only passed here so if no summary is shown, the questions can't advance.

david-rosales:
@kevinrobinson This is a great idea! I found one main issue where if showing the summary is disabled, the respond/save response/pass buttons do not proceed to the next question. I think they still submit the response but `onDone` is never called.

I think this structure would actually work well with the texting-based UI. The way it is set up inside of `MobileInterface` is already in a `ScenarioRenderer` and `ResponseRenderer` format with `TextBody` and `TextFooter`.

ui/src/message_popup/popup_question.jsx
kevinrobinson: Good catch!  I ran into something similar adding in video scenarios, and so I'll make the minimal change to fix this and then I think we'll have to rework this a bit.

This is a place where tests can help too, when there are lots of different combinations and we want to check that we don't break one when reworking another.

ui/src/message_popup/popup_question_test.jsx
```diff
@@ -0,0 +1,47 @@
...
+    scaffolding: TestFixtures.practiceScaffolding,
+    limitMs: 30000,
+    onLog: sinon.spy(),
+    onDone: sinon.spy(),
+    isLastQuestion: false,
+    ...props
+  };
+}
+
+function expectChildElementsIn(wrapper) {
```
kevinrobinson: This is an example of having repeated assertions, and wanting to factor it out to re-use in different test cases.

ui/src/message_popup/popup_question_test.jsx
```diff
@@ -0,0 +1,47 @@
...
+import * as TestFixtures from './test_fixtures.js';
+
+import PopupQuestion from './popup_question.jsx';
+import SummaryCard from './summary_card.jsx';
+import ScenarioRenderer from './renderers/scenario_renderer.jsx';
+import PromptsRenderer from './renderers/prompts_renderer.jsx';
+import ResponseRenderer from './renderers/response_renderer.jsx';
+
+
+function testProps(props) {
```
kevinrobinson: This is a common pattern - creating props for a component, and allowing variations on this to be passed in.

ui/src/message_popup/renderers/response_renderer.jsx
```diff
@@ -0,0 +1,194 @@
...
+
+    // Capture response either way
+    this.logMessage('message_popup_response');
+
+    // Transition to revising or to being done
+    if (scaffolding.helpType === 'feedback' && examples.length > 0 && !isDoneRevising) {
+      this.setState({ isRevising: true });
+    } else {
+      this.submitResponse();
+    }
```
kevinrobinson: Found a bug in here in testing - previously it wasn't logging the response in solution mode.

ui/src/message_popup/renderers/response_renderer_test.jsx
```diff
@@ -0,0 +1,70 @@
...
+import FeedbackCard from '../feedback_card.jsx';
+import Snackbar from 'material-ui/Snackbar';
+
+function testProps(props) {
+  return {
+    question: TestFixtures.testQuestion,
+    scaffolding: TestFixtures.practiceScaffolding,
+    limitMs: 30000,
+    onLogMessage: sinon.spy(),
+    onResponseSubmitted: sinon.spy(),
```
kevinrobinson: A "spy" is an object that can be called like a function, but that just tracks when it was called and with what arguments.  It can also return fake values.  Here we're using it to be able to test whether these callbacks get called when typing in responses and hitting "save response."

ui/src/message_popup/renderers/response_renderer_test.jsx
```diff
@@ -0,0 +1,70 @@
...
+  it('submits response immediately in solution mode', () => {
+    const props = testProps({ scaffolding: TestFixtures.solutionScaffolding });
+    const wrapper = shallow(<ResponseRenderer {...props} />);
+    const responseText = 'foo';
+
+    wrapper.find(TextField).simulate('change', { target: { value: responseText } });
+    wrapper.find(RaisedButton).simulate('touchTap');
+    expect(wrapper.find(FeedbackCard).length).to.equal(0);
+    expect(props.onLogMessage.callCount).to.equal(1);
+    expect(props.onResponseSubmitted.firstCall.args[0]).to.deep.equal({
```
kevinrobinson: This is checking that `ResponseRenderer` makes a call into this prop function, and that it passes back the values we'd expect.

ui/src/message_popup/test_fixtures.js
```diff
@@ -0,0 +1,20 @@
+/* flow weak */
+import _ from 'lodash';
+import {allQuestions} from './questions.js';
+import {withStudents} from './transformations.jsx';
+
+
+/*
+These are fixtures for JavaScript tests
```
kevinrobinson: The idea here is that these are only used in test, and they're split out here since we want to be able to re-use them in different test scenarios.

These will need to be updated as the code itself changes too.

kevinrobinson:
@david-rosales Thanks for the awesome review!  I added some more tests in, mind checking them out?  We can talk through them too tomorrow morning.

Also 👍  on adapting the texting UI to this structure.  My assumption is that'd be awesome to do but probably too ambitious to finish this week (given the existing work on finishing the authoring UI and adding some tests).  But let me know what you think, I'm happy to help out with that or if you think it's realistic to try that too we could go for it.  And no pressure either way! :)

ui/src/message_popup/popup_question.jsx
```diff
@@ -49,221 +47,90 @@ export default React.createClass({
...
+        this.onDone();
+      }
+    });
   },
 
-  onRequestClose() {
-    //This empty function is meant to cancel out the closing of the toast without having to use
-    //an extremely large autoHideDuration
+  onDone() {
+    this.props.onDone(this.state.response.elapsedMs / 1000);
```
david-rosales: I think `this.props.onDone` takes in milliseconds so just `this.state.response.elapsedMs` would work.

david-rosales:
@kevinrobinson Looks good, I just found one thing but other than that, approved.

ui/src/message_popup/popup_question.jsx
kevinrobinson: Good catch!  I must have fixed this on another branch or something, updated.

PR #122, opened 2016-08-15
Fail tests on console.error or PropType check, co-locate tests with source

kevinrobinson:
This adds setup code to fail tests on `console.error` (which occurs when a React PropType check fails).

This shows tests failing like they should when required PropTypes are missing:
screen shot 2016-08-15 at 6 47 56 pm

This also updates tests to be co-located with their source rather than in a separate `test` folder, with the idea being that test should be close to their code and move with it, rather than mirroring the src folder structure elsewhere.


kevinrobinson:
@david-rosales Can you review this por favor?

ui/src/message_popup/renderers/hint_card_test.jsx
```diff
@@ -45,31 +37,4 @@ describe('<HintCard />', () => {
...
-    }, 1500);
-  });
 });
-
-
-
-// nested integration-style tests
-// rendering inside VelocityTransitionGroup
-// web inspector debugging
-// warning only if prop is missing or the wrong type
```
kevinrobinson: These were my notes when initially setting this up, I pulled them out and will track and tackle them separately.

ui/test/hello_react_test.jsx
```diff
@@ -1,47 +0,0 @@
...
-    expect(wrapper.find(Foo)).to.have.length(1);
-    expect(wrapper.find('.plain-div')).not.to.have.length(1);
-  });
-
-  it('can render to plain HTML', () => {
-    const wrapper = render(<MuiThemeProvider><Foo hello="world" /></MuiThemeProvider>);
-    expect(wrapper.find('.plain-div')).to.have.length(1);
-    expect(wrapper.find('button')).to.have.length(1);
-  });
-});
```
kevinrobinson: @david-rosales This was just to demonstrate and talk through when pairing, but we can grab this from source control if we want or just start with some new real tests, no need to keep this around.

david-rosales:
@kevinrobinson Looks good, just a few flow errors that need updated imports of `hint_card.jsx`.

kevinrobinson:
@david-rosales Thanks for catching (and sorry for the sloppiness with not checking the Travis build)!  Juggling too many branches :)

PR #125, opened 2016-08-16
Initial video scenarios

kevinrobinson:
This is on top of https://github.com/mit-teaching-systems-lab/threeflows/pull/121, which should be merged first.  This is the [relevant diff](https://github.com/mit-teaching-systems-lab/threeflows/pull/125/commits/cfb6428b40bd0dc10649ee8f83e2302b40dac50f).

This adds functional video scenario questions, using YouTube videos.  It uses react-youtube to play the video, mostly to wire up a callback for when the video is done playing.  Video questions require a `youTubeId` field, and `ScenarioRenderer` uses duck typing and starts with `youTubeId` and then `text`.  We could update that to degrade or do something smarter if video wasn't available, but there's nothing more sophisticated right now.

The player settings are set to disabled as much as possible, with the exception of showing the scrubber and audio (since removing them adds more prominent YouTube branding that's visible the whole time the video is playing).  There's an `autoplay` param but it doesn't seem to work.

This change also involves some refactoring of `PopupQuestion` to support letting the scenario determine when to show the response UI.

Watching a video:
screen shot 2016-08-16 at 6 51 59 pm

Seeing the response appear after the video ends (with the video resetting to the first frame):
screen shot 2016-08-16 at 6 54 36 pm


kevinrobinson:
@david-rosales Can you check this out?  I'm also curious to try it on actual mobile (even though the Sept. playtest will be on desktop) but was going to get this working and deploy it first.

david-rosales:
@kevinrobinson Looks good and works well. I checked how it would look on my phone with ngrok and the only suggestion I have is to somehow make the height dynamic because on a smaller portrait style screen, there is a lot of black space on the top and bottom. Approved.

kevinrobinson:
@david-rosales Nice!  ngrok is a great idea.  Yeah the fixed height here isn't great - I'll try to clear out these open PRs and get them merged first, and then address that after.

PR #126, opened 2016-08-17
Functional audio responses

kevinrobinson:
This is on top of https://github.com/mit-teaching-systems-lab/threeflows/pull/125, which should be merged first.  The relevant diff is the [last commit](https://github.com/mit-teaching-systems-lab/threeflows/commit/c37fc1b8b62034f1221b26cbf5b265fd9b43ddbd).

This adds initial audio responses, using the AuO library underneath.  It only works on desktop Chrome and Firefox, which should be sufficient for the September playtest.  Medium term, we'll have to look at native options for better audio and video recording support.  The summary page after each question is not shown for audio, regardless of the settings (since we can't represent the audio response with text at the moment).

In the implementation, the major changes are to:
- move tracking elapsed time into `PopupQuestion` and out of the response components
- have `PopupQuestion` determine a `responseMode` for each question (or use a provided strategy)
- branch on response mode in `PopupQuestion#renderResponse`
- handle details of recording audio in `AudioResponse`, adapted mostly unchanged from the previous `AudioFlowPrototype`.

## Prompt to record
screen shot 2016-08-17 at 10 17 34 am

## Recording...
screen shot 2016-08-17 at 10 17 39 am

## Option to listen and re-record
screen shot 2016-08-17 at 9 51 03 am

## Audio responses for video scenarios
screen shot 2016-08-17 at 9 54 13 am


ui/src/message_popup/popup_question.jsx
```diff
@@ -118,19 +154,29 @@ export default React.createClass({
...
           question={question}
         />
-        {showResponse && <ResponseRenderer
-          scaffolding={scaffolding}
-          question={question}
-          limitMs={limitMs}
-          onResponseSubmitted={this.onResponseSubmitted}
-          onLogMessage={this.logData}
-        />}
+        {allowResponding && this.renderResponse()}
```
david-rosales: For some reason, when the regular text response is rendered, the style is broken. The style gets fixed when the user changes the text in the textfield.

screen shot 2016-08-17 at 3 57 11 pm
vs.
screen shot 2016-08-17 at 3 57 21 pm

I ran into a similar issue earlier with Question Authoring but I'm not too sure if it is the same. I was updating the state that belonged in the textfield shortly after it was created and even though the textfield had a paragraph of text, it showed as a single line until edited. Even if it is the same problem, I just changed the code so that I wouldn't have to update the state so I never really found the solution. It's possible that it is also an issue with Material UI but I'm not sure.

david-rosales:
@kevinrobinson Interestingly, there is still a range of recording length where the recording fails. I still haven't been able to pinpoint it but it's somewhere around 7~10 seconds. 
```
AuO.js:798 Uncaught TypeError: Cannot read property 'length' of null
```
I went through longer recordings and those never failed so I have no clue what could be causing it.

Also, the first time a submitted recording gets saved, the following error pops up.
```
Warning: Failed prop type: Required prop `responseText` was not specified in `SummaryCard`.
```

ui/src/message_popup/popup_question.jsx
kevinrobinson: Yeah thanks for pointing this out!  I noticed this earlier but it went away.  Poking around now but no luck yet, so I'll maybe spend a little more time and if not workaround for now by forcing it to be a particular size with styles or something slightly hacky like that.

kevinrobinson:
@david-rosales Yeah thanks!  The `responseText` bug was because it's only really for text responses anyway and previously was expressed implicitly by calling `onDone` when the audio response was submitted (which would cause the parent component to update state, re-render and then unmount the <PopupQuestion /> component.  I updated `onResponseSubmitted` to call `onDone` first, and then added another guard in `renderText` to be more explicit too.

For the longer recordings, I have that next on my list but didn't tackel it yet.  I should have called that out in the pull request description, thanks for making sure it gets fixed :)

ui/src/message_popup/popup_question.jsx
kevinrobinson: This is coming from the interaction between `VelocityTransitionGroup` in the experience page, and the Material UI `TextField`.  Since this is being animated, the root `TextField` element is mounted into the DOM before it is visible and has its full size.  The `EnhancedTextarea` inside the `TextField` component looks at the DOM to see how big it should be, and sizes itself based on that (presumably to lay out the different hint and float text labels just so).  On the initial render, this code looks at the DOM and sees it has 0 height since it's not visible yet.  After the first text is entered, `TextField` runs its layout code again and then realizes it can take up more space since the element has height now, and sizes it correctly.

I worked around this by fixing the height of the `TextField` component to be correct for two rows heigh at this font size.  In general, animating components in and out often means you have to fix some parts of the layout or sizing, so while this isn't great it seems like a reasonable workaround.  The only real cost is having to update the sizing and number of rows, and I called that out with a comment here.

kevinrobinson:
@david-rosales I worked around the long audio recording issue.  This was happening because after AuO stopped recording, it does some post-processing before the audio buffer is available.  To account for this, it disables the UI while this happens.  The hacked code I wrote here called "stop" and the "save" directly, and so it was calling save and trying to access the audio buffer before the post-processing was done.  The workaround here hacks in a callback to AuO so the the application code properly waits until the post-processing is done before calling `save` to get the blob.

@wqian94 FYI, and thanks again for letting us hack around with your great project! 👍

ui/.babelrc
```diff
@@ -13,5 +13,6 @@
         }]
       ]
     }
-  }
+  },
+  "compact": false
```
kevinrobinson: This is working around a Babel warning about Au0 being >100k.

PR #127, opened 2016-08-17
Add querystring showing UI to allow forcing modes

kevinrobinson:
This adds a query string param that will show a UI allowing the user to set the strategy for picking audio or text responses.  For now it's either randomly mixed or all of a particular response type.

I also spent some time trying to get full renders working during test.  This involved:
- making a new `TestAuthContainer` to inject authentication context for test
- adding a call to `react-tap-event-plugin` in the Mocha `setup.js`
- running into a Material UI component relying on `element.scrollHeight`, which isn't supported in `jsdom`
- filing [this workaround upstream](https://github.com/callemall/material-ui/pull/5015) that I don't expect will be merged, and don't have any great ideas to workaround
- adding a minimal static HTML render test

### URL:
screen shot 2016-08-17 at 3 15 00 pm

### UI:
screen shot 2016-08-17 at 3 14 55 pm

### DIFF
--- a/ui/src/message_popup/experience_page.jsx
+++ b/ui/src/message_popup/experience_page.jsx
@@ -151,7 +151,7 @@ export default React.createClass({
 
   renderDone() {
     return (
-      <div>
+      <div className="done">
         <VelocityTransitionGroup enter={{animation: "slideDown"}} leave={{animation: "slideUp"}} runOnMount={true}>
           <div style={_.merge(_.clone(styles.container), styles.done)}>
             <div style={styles.doneTitle}>You finished!</div>
@@ -176,7 +176,7 @@ export default React.createClass({
 
     return (
       <VelocityTransitionGroup enter={{animation: "callout.pulse", duration: 500}} leave={{animation: "slideUp"}} runOnMount={true}>
-        <div>
+        <div className="instructions">
           <InstructionsCard 
            itemsToShow={query}
            />
@@ -199,7 +199,7 @@ export default React.createClass({
     const question = questions[questionsAnswered];
 
     return (
-      <div style={styles.container}>        
+      <div className="question" style={styles.container}>        
         <LinearProgress color="#EC407A" mode="determinate" value={questionsAnswered} max={sessionLength} />
         <VelocityTransitionGroup enter={{animation: "slideDown"}} leave={{animation: "slideUp"}} runOnMount={true}>
           <PopupQuestion
@@ -228,7 +228,7 @@ export default React.createClass({
     const sessionLength = questions.length;
     const question = withStudents(questions)[questionsAnswered];
     return ( 
-      <div>
+      <div className="prototype">
         <LinearProgress color="#EC407A" mode="determinate" value={questionsAnswered} max={sessionLength} />
         <MobileInterface
           key={JSON.stringify(question)}
--- a/ui/src/message_popup/experience_page.jsx
+++ b/ui/src/message_popup/experience_page.jsx
@@ -98,13 +98,15 @@ export default React.createClass({
     Routes.navigate(Routes.Home);
   },
   
-  onSaveScaffoldingAndSession(scaffolding, email, questions){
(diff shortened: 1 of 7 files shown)


ui/src/message_popup/experience_page.jsx
```diff
@@ -129,16 +125,33 @@ export default React.createClass({
...
         />
-        {!hasStarted && this.renderInstructions()}
-        {hasStarted && questionsAnswered >= sessionLength && this.renderDone()}
-        {hasStarted && questionsAnswered < sessionLength && (_.has(this.props.query, 'mobilePrototype') ? this.renderMobilePrototype() : this.renderPopupQuestion())}
+        {this.renderMainScreen()}
       </div>
     );
   },
 
+  renderMainScreen() {
```
kevinrobinson: Unrelated cleanup

ui/src/message_popup/experience_page.jsx
```diff
@@ -159,16 +172,19 @@ export default React.createClass({
 
   renderInstructions() {
     const {scaffolding} = this.state;
+    const {query} = this.props;
+
     return (
       <VelocityTransitionGroup enter={{animation: "callout.pulse", duration: 500}} leave={{animation: "slideUp"}} runOnMount={true}>
-        <div>
+        <div className="instructions">
```
kevinrobinson: @david-rosales This isn't ideal but since I couldn't get the `mount` deep renderer working yet, this is how you can add in classNames to check that the output of render is what you'd expect.

ui/src/message_popup/experience_page_test.jsx
```diff
@@ -0,0 +1,38 @@
...
+    </MuiThemeProvider>
+  );
+}
+
+// TODO(kr) it doesn't yet work to do a full render of ExperiencePage,
+// because the TextField Material UI component relies on element.scrollHeight,
+// which isn't present in jsdom.
+// Not sure of a good workaround, but submitted
+// https://github.com/callemall/material-ui/pull/5015 as a workaround and
+// to get input.
```
kevinrobinson: 😞

kevinrobinson:
@david-rosales Can you review por favor?

ui/src/message_popup/experience_page_test.jsx
kevinrobinson: yay for oviertassinari 😄

ui/src/message_popup/scaffolding_card.jsx
```diff
@@ -99,7 +119,7 @@ export default React.createClass({
   },
   
   render(){
-    const {itemsToShow} = this.props;
+    const {itemsToShow, allowChoosingResponseMode} = this.props;
```
david-rosales: I wonder if we should make it so that `allowChoosingResponseMode` is enabled when the query string 'all' is present as well.

ui/src/message_popup/experience_page.jsx
```diff
@@ -159,16 +172,19 @@ export default React.createClass({
...
           <InstructionsCard 
-           itemsToShow={this.props.query}
+           itemsToShow={query}
            />
           <ScaffoldingCard
             initialEmail={this.context.auth.userProfile.email}
             scaffolding={scaffolding}
-            itemsToShow={this.props.query}
+            itemsToShow={query}
+            allowChoosingResponseMode={_.has(query, 'modes')}
```
david-rosales: Could we handle this inside of `ScaffoldingCard` because `itemsToShow` already contains `query`?

david-rosales:

ui/src/message_popup/scaffolding_card.jsx
kevinrobinson: Cool, that makes sense with moving the check into `ScaffoldingCard`, updated!

kevinrobinson:
@david-rosales Thanks for the great feedback and review! 👍

PR #128, opened 2016-08-18
Adds tests for QuestionsPage

david-rosales:
I added a few tests for `QuestionsPage` that look for rendering when loaded, not loaded, and with no questions available.

This was built on top of #118 so that should get merged first. I also think I want to wait for #127 to get merged first because I want to see if I can use the `TestAuthContainer` to test some more things on `QuestionsPage`.

### DIFF
--- a/ui/src/message_popup/author/questions_page_test.jsx
+++ b/ui/src/message_popup/author/questions_page_test.jsx
@@ -18,27 +18,19 @@ function testProps(props) {
 }
 
 describe('<QuestionsPage />', ()=>{
-  it('renders current questions when loaded', ()=>{
+  it('renders questions when loaded', ()=>{
     const props = testProps();
     const wrapper = shallow(<QuestionsPage {...props} />);
     expect(wrapper.find(QuestionButton)).to.not.have.length(0);
-  });
-  it('does not render current questions when not loaded', ()=>{
-    const props = testProps({loaded: false});
-    const wrapper = shallow(<QuestionsPage {...props} />);
-    expect(wrapper.find(QuestionButton)).to.have.length(0);
-  });
-  it('renders archived questions when loaded', ()=>{
-    const props = testProps();
-    const wrapper = shallow(<QuestionsPage {...props} />);
     expect(wrapper.find(ArchivedQuestionButton)).to.not.have.length(0);
   });
-  it('does not render archived questions when not loaded', ()=>{
+  it('does not render questions when not loaded', ()=>{
     const props = testProps({loaded: false});
     const wrapper = shallow(<QuestionsPage {...props} />);
+    expect(wrapper.find(QuestionButton)).to.have.length(0);
     expect(wrapper.find(ArchivedQuestionButton)).to.have.length(0);
   });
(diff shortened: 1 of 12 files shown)


ui/src/message_popup/author/questions_page_test.jsx
```diff
@@ -0,0 +1,61 @@
...
+    const props = testProps({loaded: false});
+    const wrapper = shallow(<QuestionsPage {...props} />);
+    expect(wrapper.find(QuestionButton)).to.have.length(0);
+  });
+  it('renders archived questions when loaded', ()=>{
+    const props = testProps();
+    const wrapper = shallow(<QuestionsPage {...props} />);
+    expect(wrapper.find(ArchivedQuestionButton)).to.not.have.length(0);
+  });
+  it('does not render archived questions when not loaded', ()=>{
```
kevinrobinson: It might simplify to lump this case and the "does not render current questions when not loaded" case into the same `it` block, up to you.

ui/src/message_popup/author/questions_page_test.jsx
```diff
@@ -0,0 +1,61 @@
...
+    loaded: true,
+    ...props
+  };
+}
+
+describe('<QuestionsPage />', ()=>{
+  it('renders current questions when loaded', ()=>{
+    const props = testProps();
+    const wrapper = shallow(<QuestionsPage {...props} />);
+    expect(wrapper.find(QuestionButton)).to.not.have.length(0);
```
kevinrobinson: Would it make sense to assert on how many questions there are, or that the number rendered matches the number in `testQuestions `?

ui/src/message_popup/author/questions_page_test.jsx
```diff
@@ -0,0 +1,61 @@
...
+    const props = testProps({
+      allQuestions: {
+        currentQuestions: [],
+        archivedQuestions: []
+      }
+    });
+    const wrapper = shallow(<QuestionsPage {...props} />);
+    expect(wrapper.find(QuestionButton)).to.have.length(0);
+  });
+  it('does not render archived questions when none exist', ()=>{
```
kevinrobinson: You could also put the assertion here in the same case with the "does not render current questions when none exist", into a test case that checks two `expects`.

kevinrobinson: This is mostly just about what seems simpler and more direct and up to you.  The only real tradeoff is that if there is one `it` block with two assertions, that is printed out as a single test case in the test output.  And that if the first assertion fails, the second one won't be run.

kevinrobinson:
@david-rosales The tests here look great, a few minor style suggestions for you and 🚢 whenever that other work clears.

david-rosales:
@kevinrobinson Thanks for the feedback, I'll merge this after we get the other work merged.

david-rosales:
@kevinrobinson I actually think these tests would be very helpful with changing #118 to have the question loading back in the individual components. I'm going to merge this branch into #118 for now.

kevinrobinson:
@david-rosales sounds good! 👍

PR #129, opened 2016-08-18
Fix for if the questions table has no rows

david-rosales:
Changed: 
```
if(rows.length === 0) return response.json({});
```
to:
```
if(rows.length === 0) return response.json({questions: {currentQuestions:[], archivedQuestions: []}});
```


kevinrobinson:
Approved.

david-rosales:
@kevinrobinson This is a small fix, can you review?

PR #130, opened 2016-08-19
Add note about matching node version

kevinrobinson:
Thanks @kesiena115!


kevinrobinson:
Self-approving this, we talked about it in person.  cc @kesiena115

PR #131, opened 2016-08-19
Reorganize docs to help first-time setup, and add in notes about S3

kevinrobinson:
This was from feedback after @kesiena115 used this to get set up.


kevinrobinson:
@kesiena115 Can you review and comment on anything else you think we can improve here?  Thanks!

kevinrobinson:
@kesiena115 Merging this, if you're able to review after-the-fact that'd be awesome but no worries if not.

PR #132, opened 2016-08-19
Fix bug with not logging audio responses

kevinrobinson:
The audio files are captured, but the response records weren't being logged.  Adds comments as well indicating that response components need to log all their own data.

### DIFF
--- a/ui/src/message_popup/renderers/audio_response.jsx
+++ b/ui/src/message_popup/renderers/audio_response.jsx
@@ -20,11 +20,13 @@ export default React.createClass({
     question: React.PropTypes.object.isRequired,
     scaffolding: React.PropTypes.object.isRequired,
     limitMs: React.PropTypes.number.isRequired,
+    elapsedMs: React.PropTypes.number.isRequired,
     onLogMessage: React.PropTypes.func.isRequired,
     onResponseSubmitted: React.PropTypes.func.isRequired
   },
 
   onDone(audioUrl) {
+    this.props.onLogMessage('message_popup_audio_response', {audioUrl});
     this.props.onResponseSubmitted({audioUrl});
   },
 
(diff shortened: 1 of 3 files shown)


kevinrobinson:
@david-rosales Are you able to review later right before dropping the 🎤 ?

ui/src/message_popup/renderers/audio_response.jsx
```diff
@@ -20,11 +20,13 @@ export default React.createClass({
     question: React.PropTypes.object.isRequired,
     scaffolding: React.PropTypes.object.isRequired,
     limitMs: React.PropTypes.number.isRequired,
+    elapsedMs: React.PropTypes.number.isRequired,
```
david-rosales: Was this necessary or will it be needed later?

ui/src/message_popup/renderers/audio_response_test.jsx
```diff
@@ -0,0 +1,44 @@
...
+import * as TestFixtures from '../test_fixtures.js';
+
+import AudioResponse from './audio_response.jsx';
+import AudioRecorderFlow from '../../components/audio_recorder_flow.jsx';
+
+function testProps(props) {
+  return {
+    question: TestFixtures.testQuestion,
+    scaffolding: TestFixtures.practiceScaffolding,
+    limitMs: 30000,
```
david-rosales: Same as the above comment

ui/src/message_popup/renderers/audio_response_test.jsx
```diff
@@ -0,0 +1,44 @@
...
+    elapsedMs: 0,
```
david-rosales: Same as the above comment

david-rosales:

ui/src/message_popup/renderers/audio_response.jsx
kevinrobinson: Yeah I added it in since it's being passed in now that `PopupQuestion` manages that time and uses the same set of props with each response component.  I think it's doesn't super matter, but this seems slightly better since `elapsedMs` is information that is available to a response component (and the audio component just happens not to use it).  Could go either way though.

PR #133, opened 2016-08-19
Copy revisions for response modes, additional tests

kevinrobinson:
This makes some small copy changes to emphasize user actions rather than technology or medium, and also adds some new tests.  In the process, I discovered mistakes in flow annotations for existing test files (`flow weak` instead of `@flow weak`) and added those back.

Before:
screen shot 2016-08-19 at 2 31 48 pm

After:
screen shot 2016-08-19 at 2 41 37 pm

### DIFF
--- a/ui/src/message_popup/scaffolding_card_test.jsx
+++ b/ui/src/message_popup/scaffolding_card_test.jsx
@@ -1,0 +1,59 @@
+/* @flow weak */
+import React from 'react';
+
+import {shallow} from 'enzyme';
+import sinon from 'sinon';
+import {expect} from 'chai';
+import * as TestFixtures from './test_fixtures.js';
+
+import ScaffoldingCard from './scaffolding_card.jsx';
+import RaisedButton from 'material-ui/RaisedButton';
+import {RadioButtonGroup} from 'material-ui/RadioButton';
+import TextField from 'material-ui/TextField';
+import Toggle from 'material-ui/Toggle';
+import Slider from 'material-ui/Slider';
+
+
+function testProps(props) {
+  return {
+    scaffolding: TestFixtures.practiceScaffolding,
+    initialEmail: 'user@test.com',
+    query: {},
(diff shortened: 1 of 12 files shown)


kevinrobinson:
@david-rosales This has Flow errors, which I'll fix later today but if you get a chance check out the different selector syntax that Enzyme supports for checking for components with particular props.  It's like CSS selectors but React-y.

ui/src/message_popup/scaffolding_card_test.jsx
```diff
@@ -0,0 +1,56 @@
...
+    initialEmail: 'user@test.com',
+    query: {},
+    onSessionConfigured: sinon.spy(),
+    ...props
+  };
+}
+
+function expectCoreOptions(wrapper) {
+  expect(wrapper.text()).to.contain('Practice scenarios:');
+  expect(wrapper.find('RadioButtonGroup[name="indicatorId"]').length).to.equal(1);
```
david-rosales: Ohh that's useful!

david-rosales:
@kevinrobinson I tried adding a flow annotation to a test file before and it would not recognize `describe` and `it`. 

After looking into it, I found that adding this somewhere in the test will fix that:
```
declare function describe(name:string, callback:Function):void;
declare function it(name:string, callback:Function):void;
```

This is where I found that: https://github.com/facebook/flow/issues/1632

ui/src/message_popup/scaffolding_card_test.jsx
kevinrobinson: yeah, just noticed it!  https://github.com/airbnb/enzyme/blob/master/docs/api/selector.md

kevinrobinson:
@david-rosales Thanks for being so helpful!  🎉  I added in those declarations and worked around the `TestAuthContainer` error (which I think may be a Flow bug, possibly related to https://github.com/facebook/flow/issues/1355 or https://github.com/facebook/flow/issues/1964).

PR #134, opened 2016-08-19
Adds Authoring UI component tests

david-rosales:
I added a few tests for the individual components that make up the authoring UI.

I was originally going to make tests for `EditQuestionPage` and `NewQuestionPage` but that is currently being blocked by the fact that `mount` currently has an issue with material-ui textfields.

### DIFF
--- a/ui/src/message_popup/author/question_editing_components/indicators_test.jsx
+++ b/ui/src/message_popup/author/question_editing_components/indicators_test.jsx
@@ -7,7 +7,7 @@ import Indicators from './indicators.jsx';
 
 import {RadioButton, RadioButtonGroup} from 'material-ui/RadioButton';
 
-import {testQuestion} from '../../test_fixtures.js'
+import {testQuestion} from '../../test_fixtures.js';
 import {withIndicator} from '../../transformations.jsx';
 import {indicators} from '../../../data/indicators.js';
 
--- a/ui/src/message_popup/author/question_editing_components/indicators_test.jsx
+++ b/ui/src/message_popup/author/question_editing_components/indicators_test.jsx
@@ -1,0 +1,44 @@
+import React from 'react';
+import {shallow} from 'enzyme';
+import {expect} from 'chai';
+import sinon from 'sinon';
+
+import Indicators from './indicators.jsx';
+
+import {RadioButton, RadioButtonGroup} from 'material-ui/RadioButton';
+
+import {testQuestion} from '../../test_fixtures.js'
+import {withIndicator} from '../../transformations.jsx';
+import {indicators} from '../../../data/indicators.js';
+
+function testProps(props){
+  return ({
+    indicator: withIndicator(testQuestion).indicator,
+    onIndicatorChange: sinon.spy()
+  });
+}
+
+describe('<EditingComponent.Inidicators />', ()=>{
+  it('renders all the available indicators', ()=>{
+    const props = testProps();
+    const wrapper = shallow(<Indicators {...props} />);
+    expect(wrapper.find(RadioButtonGroup)).to.have.length(1);
+    expect(wrapper.find(RadioButton)).to.have.length(indicators.length);
+    wrapper.find(RadioButton).forEach((node) => {
+      expect(node.props().value).to.be.oneOf(indicators.map(indicator => indicator.id.toString()));
+    });
+  });
+  it('selects the correct initial indicator', ()=>{
+    const props = testProps();
+    const wrapper = shallow(<Indicators {...props} />);
+    expect(wrapper.find(RadioButtonGroup).props().valueSelected).to.equal(testQuestion.indicatorId.toString());
+  });
+  it('changes to the correct indicator', ()=>{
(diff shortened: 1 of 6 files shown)


david-rosales:
@kevinrobinson Can you review this when you get a chance?

ui/src/message_popup/author/question_editing_components/indicators_test.jsx
```diff
@@ -0,0 +1,44 @@
...
+    const wrapper = shallow(<Indicators {...props} />);
+    expect(wrapper.find(RadioButtonGroup).props().valueSelected).to.equal(testQuestion.indicatorId.toString());
+  });
+  it('changes to the correct indicator', ()=>{
+    const props = testProps();
+    const wrapper = shallow(<Indicators {...props} />);
+    const newIndicator = indicators[1];
+    wrapper.find(RadioButtonGroup).simulate('change', {target: {value: newIndicator.id}});
+    expect(props.onIndicatorChange.callCount).to.equal(1);
+    expect(props.onIndicatorChange.firstCall.args).to.deep.equal([{target: {value: newIndicator.id}}]);
```
kevinrobinson: Nice!  These are great examples of low-level unit tests, when you want to really thoroughly testing every aspect of a component.

ui/src/message_popup/author/question_editing_components/indicators_test.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+describe('<EditingComponent.Inidicators />', ()=>{
+  it('renders all the available indicators', ()=>{
+    const props = testProps();
+    const wrapper = shallow(<Indicators {...props} />);
+    expect(wrapper.find(RadioButtonGroup)).to.have.length(1);
+    expect(wrapper.find(RadioButton)).to.have.length(indicators.length);
+    wrapper.find(RadioButton).forEach((node) => {
+      expect(node.props().value).to.be.oneOf(indicators.map(indicator => indicator.id.toString()));
+    });
```
kevinrobinson: Nice!  One wrinkle with this approach is that the test will still pass if say the first indicator repeated n times.  That's not likely to be a real bug, but you could also map over all the `RadioButton` components and get their `props().value` values, and then compare that to the set of indicatorIds and assert they're equal.  It doesn't matter all that much here, but that's maybe slightly better for other circumstances like this.

ui/src/message_popup/author/question_editing_components/question_text_test.jsx
```diff
@@ -0,0 +1,41 @@
...
+    const props = testProps({originalText: undefined});
+    const wrapper = shallow(<QuestionText {...props} />);
+    expect(wrapper.find(".originalQuestionText")).to.have.length(0);
+  });
+  it('returns the changed text', ()=>{
+    const props = testProps();
+    const wrapper = shallow(<QuestionText {...props} />);
+    wrapper.find(TextField).simulate('change', { target: { value: testQuestion.text } });
+    expect(props.onQuestionTextChange.callCount).to.equal(1);
+    expect(props.onQuestionTextChange.firstCall.args).to.deep.equal([{ target: { value: testQuestion.text } }]);
```
kevinrobinson: This is a minor point I should have noticed earlier, but I think in general it'd be better to pass just `testQuestion.text` to the callback - no need to pass it the full DOM event.  So concretely the `QuestionText` component wouldn't directly pass in `props.onQuestionTextChange` to `TextField` but would pass its own method, and then grab the value from the event and pass only that on `props.onQuestionTextChange`.

This is true for the other components that `QuestionPage` uses too.

ui/src/message_popup/author/questions_page_test.jsx
```diff
@@ -15,6 +19,16 @@ function testProps(props) {
   };
 }
 
+function withContext(child) {
```
kevinrobinson: Awesome!  This is from `experience_page_test` and seems generally useful, so I'd pull it out into another file and the `test` folder is probably the right home for it.

kevinrobinson:
@david-rosales This looks great, just some minor style suggestions, approved!

If you'd like to finish this off feel free (sorry for not getting to this until now), but definitely don't feel like you need to.  Otherwise I'll take it from here and merge this Monday afternoon.

Thanks, it's been awesome working with you this summer! 😄

ui/src/message_popup/author/question_editing_components/question_text_test.jsx
kevinrobinson: I did this now.  I also ran into some problems with React invariant violations around `onSetChildren` and `onSetParent` when hot-patching tests with mocha-watch.  I suspected this was from calling `Routes.navigate` which I was surprised ever works.  I ended up mocking those out and also removing the mounting test in `questions_page_test` since these both would lead to the invariant.

This works around, but it'll be good to spend some time and get tests working that can do normal mounting, either with Enzyme or without.

kevinrobinson:
Moving to https://github.com/mit-teaching-systems-lab/threeflows/pull/136.

PR #135, opened 2016-08-23
Add feature detection for AudioCapture

kevinrobinson:
Feature detect `getUserMedia` as part of drawing a response mode for a question, and fall back to text if unsupported.


kevinrobinson:
image

PR #136, opened 2016-08-24
Adds authoring UI component tests

kevinrobinson:
Finishing up the remaining bits from @david-rosales' awesome work here: https://github.com/mit-teaching-systems-lab/threeflows/pull/134.

PR #137, opened 2016-08-24
Copy tweaks to authoring UI

kevinrobinson:
These were described in https://github.com/mit-teaching-systems-lab/threeflows/pull/136#issuecomment-242219338 but I mistakenly left out that last commit.


kevinrobinson:
selfie

PR #139, opened 2016-08-29
Replacing AuO and addressing dropped audio frames

kevinrobinson:
This removes AuO and replaces it with a more minimal implementation that uses the `getUserMedia` API directly.  It's adapted from https://github.com/danrouse/react-audio-recorder and copies the WAV encoding function directly from that project.

There's no user-facing changes here, other than that the audio quality is better (primarily by increasing the buffer size, since we don't care about latency to fill the buffer while recording, only quality and size of the recording when finished).


kevinrobinson:
FYI @wqian94, thanks again for sharing your awesome work!

PR #140, opened 2016-08-30
When revising text response, animate down to feedback

kevinrobinson:
On a mobile screen, the feedback may or may not be visible on the screen.  This animates scrolling down to it when it appears.


kevinrobinson:
selfie

PR #142, opened 2016-08-30
Revise video scenario height, and add scroll to response once the vid…

kevinrobinson:
…eo finishes

Shown in iPhone 5 screen:
screen shot 2016-08-30 at 3 44 24 pm


kevinrobinson:
selfie

PR #145, opened 2016-10-17
Clarify steps for installing node.js

kevinrobinson:
@kesiena115 adding in what we just talked about in person.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,11 +16,17 @@ This builds artifacts and places them in the `ui/build` folder.
 
 
 #### Server
+The server app was developed for the [engine specified in the root package.json](https://github.com/mit-teaching-systems-lab/threeflows/blob/master/package.json#L5).  If you already have node installed, you can check the version with:
+
+```
+$ node -v
+```
+And if you need to install it from scratch, follow the instructions for the current version on the [node.js website](https://nodejs.org/en/).
+
 For storing audio files, you'll need keys authorized for a development S3 bucket.  Ask someone for these.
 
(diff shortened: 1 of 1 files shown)


kevinrobinson:
@kesiena115 Can you review and confirm this makes sense?  Thanks!

kevinrobinson:
@kesiena115 Can you review this?  Thanks!

README.md
```diff
@@ -16,11 +16,17 @@ This builds artifacts and places them in the `ui/build` folder.
...
 
 #### Server
+The server app was developed for the [engine specified in the root package.json](https://github.com/mit-teaching-systems-lab/threeflows/blob/master/package.json#L5).  If you already have node installed, you can check the version with:
+
+```
+$ node -v
+```
+And if you need to install it from scratch, follow the instructions for the current version on the [node.js website](https://nodejs.org/en/).
+
 For storing audio files, you'll need keys authorized for a development S3 bucket.  Ask someone for these.
```
kesiena115: It's not clear where this key needs to be configured and now I'm guessing audio recording isn't supposed to work locally until I do this?

kesiena115:
@kevinrobinson  this looks good. I added a comment which isn't particularly related to this change but is located one line below your change.

README.md
kevinrobinson: Yeah good point!  I'll push another commit to clarify.  The server won't start if you don't have these set in a local config file that isn't checked into source: https://github.com/mit-teaching-systems-lab/threeflows/blob/master/server/s3_client.js#L4

PR #147, opened 2016-10-18
Add example of testing a method in a component in isolation

kevinrobinson:
@kesiena115 Here's an example you can look to for how to add a unit test for a particular method in a component.

If the logic or code is involved, another alternative is to split it out to a plain JS function or class and test it outside of the React component altogether.  But sometimes this is handy (as in the case where you have a method that's implementing a new strategy for choosing text and audio responses).

### DIFF
--- a/ui/src/auth_container_test.jsx
+++ b/ui/src/auth_container_test.jsx
@@ -1,0 +1,20 @@
+/* @flow weak */
+import React from 'react';
+
+import {shallow} from 'enzyme';
+import {expect} from 'chai';
+
+import AuthContainer from './auth_container.jsx';
+
+describe('<AuthContainer />', () => {
+  // Mock localStorage
+  beforeEach(() => {
+    window.localStorage = { getItem(key) { return null; } };
+  });
+
+  it('validateEmail', () => {    
+    const wrapper = shallow(<AuthContainer><div /></AuthContainer>);
+    expect(wrapper.instance().validateEmail('krob@mit.edu')).to.equal(true);
+    expect(wrapper.instance().validateEmail('krob@@@xyz')).to.equal(false);
+  });
+});


kevinrobinson:
The original test failure here was unrelated and due to a problem with jsdom not being an explicit dependency.  I'm not sure how that happened, but it's resolved in https://github.com/mit-teaching-systems-lab/threeflows/pull/148 and https://github.com/mit-teaching-systems-lab/threeflows/pull/149, which also swap in yarn for faster more explicit dependency management throughout.

kevinrobinson:
@kesiena115 Let me know if this makes sense and looks good, or if you want to talk it out too.  Can you approve when you've had a chance to check it out and confirm this makes sense?  Thanks!

kevinrobinson:
@kesiena115 Can you review this?  Thanks!

kesiena115:
@kevinrobinson sorry I don't really understand what this does or where it gets called. Perhaps a brief explanation would be helpful

kevinrobinson:
@kesiena115 Sure!  This is showing an example for how to write a unit test that tests and individual method within a React component.  This came up in some of the pairing we did a few weeks ago, where you were writing the [drawResponseMode method](https://github.com/mit-teaching-systems-lab/threeflows/blob/master/ui/src/message_popup/playtest/playtest_experience_page.jsx#L78) and were like "man, I wish we had a way to test this or verify that it's right before we just ship it!" :)

Most of the other tests in the codebase test how components work - if you give them props, do they render what we expect?  or if you render them and click on things, do the components update like we expect?  This adds a unit test that shows how to test just the logic of a single method within a component.  [This](https://github.com/mit-teaching-systems-lab/threeflows/pull/147/files#diff-4088c8aacecad129025946d9083219dfR15) is testing the validateEmail method in the [AuthContainer](https://github.com/mit-teaching-systems-lab/threeflows/blob/master/ui/src/auth_container.jsx#L89) component.  There's no particular reason I chose this component, just seemed like the kind of thing that would a good example of something that would be well-tested by an individual unit test.  This is something you could use as a model for writing a unit test for the `drawResponseMode` method.

As for where this gets called, the `describe` and `it` functions are part of [Mocha](https://mochajs.org/).  If you're more familiar with JUnit, [this article](http://www.mscharhag.com/java/oleaster-jasmine-junit-tests) has a comparison of how this would map to a JUnit test.  The Mocha test runner collects all of the files that end with `_test` and then runs all the different test cases they describe.  So this test code doesn't get called directly, but it will be run when you run the tests [like this](https://github.com/mit-teaching-systems-lab/threeflows#run-mocha-tests).

@kesiena115 Let me know if that helps!

kesiena115:
@kevinrobinson cool. This looks good to me.

kevinrobinson:
@kesiena115 👍

PR #148, opened 2016-10-18
Fix dependency bug, swap in yarn

kevinrobinson:
Addresses problem in https://travis-ci.org/mit-teaching-systems-lab/threeflows/builds/168603488, which showed jsdom not in package.json.  Not sure how these ever passed then, but this fixes the dependencies.

It also adds in yarn.lock and attempts to swap yarn in for Travis and Heroku.


kevinrobinson:
yarn looks good in Travis: https://travis-ci.org/mit-teaching-systems-lab/threeflows/builds/168608999#L329, although installing adds ~30 seconds to the build.  Good enough for now and as it's supported in Travis we can swap that in.

PR #149, opened 2016-10-18
Install UI dependencies with yarn in Heroku deploy

kevinrobinson:
Trying to yarn all the things to keep dev consistent with production and CI, etc.


kevinrobinson:
selfie

PR #152, opened 2016-10-18
Add root yarn.lock

kevinrobinson:
I was trying to test just the UI install first, but looks like this is needed to trigger the buildpack.


kevinrobinson:
selfie

PR #154, opened 2016-10-18
Revert switching over to yarn, unable to resolve brace-expansion issu…

kevinrobinson:
…e in Heroku

I haven't figured this out yet, and don't want to block other development for the playtest in the meantime.


kevinrobinson:
selfie, trying to resolve dependency problems that are only coming up in the Heroku deploy

PR #155, opened 2016-10-18
Guard stopping stream, and add call to destroy audio context

kevinrobinson:
Addressing resource leak:
```
NotSupportedError: Failed to construct 'AudioContext': The number of hardware contexts provided (6) is greater than or equal to the maximum bound (6).
```


kevinrobinson:
@kesiena115 approved in person.

PR #158, opened 2016-10-24
Add demo path

kevinrobinson:
Adds `/teachermoments/demo` that simplifies the flow and shows only a few scenarios across different representations and approximations.

screen shot 2016-10-24 at 5 36 52 pm

screen shot 2016-10-24 at 5 29 19 pm

screen shot 2016-10-24 at 5 29 24 pm

...

screen shot 2016-10-24 at 5 29 58 pm


kevinrobinson:
selfie

PR #159, opened 2016-10-26
Add some student cards to demo pages

kevinrobinson:
This was turned off before in the demo mode, show this in a few scenarios.


kevinrobinson:
selfie

PR #161, opened 2016-10-31
Revise feedback on demo, update student model

kevinrobinson:
screen shot 2016-10-31 at 3 53 56 pm

screen shot 2016-10-31 at 4 15 05 pm


kevinrobinson:
selfie cc @kesiena115

PR #163, opened 2016-11-01
Allow auth to not require email

kevinrobinson:
For gradual onboarding in demos or talks (versus playtests, where this is required up front).


kevinrobinson:
selfie

PR #165, opened 2016-11-01
Fall back to image instead of 3D model if slow network

kevinrobinson:
Fall back to a static image if Audio isn't supported (ie., it's mobile web) and network may be slow.  There are lots of other ways to improve this, but this is for the demo in a few hours where internet connectivity is limited and poor, trying to address:

screen shot 2016-11-01 at 10 49 48 am


kevinrobinson:
selfie

PR #166, opened 2016-11-01
Update demo page to remove full menu

kevinrobinson:
The navigation isn't helpful for demos or playtests.

screen shot 2016-11-01 at 10 55 39 am


kevinrobinson:
selfie

PR #170, opened 2016-11-04
Fix merge bug with AuthContainer prop

kevinrobinson:
Fixing https://travis-ci.org/mit-teaching-systems-lab/threeflows/builds/173367344, introduced from a merge conflict as another newer PR added a required prop.


kevinrobinson:
selfie

PR #173, opened 2016-12-01
Add notes about representations and approximations currently supported

kevinrobinson:
To make this more visible than just in code and project docs.


kevinrobinson:
selfie

PR #174, opened 2016-12-06
Fix bug in ResponsiveFrame scrolling on desktop

kevinrobinson:
This affected large pages on the desktop mode.

Before:
screen shot 2016-12-06 at 1 17 36 pm

After:
screen shot 2016-12-06 at 1 15 02 pm

PR #175, opened 2016-12-06
Add Lori Danson Simulation for Playtest

kesiena115:
* Add endpoint and Experience Page for Lori Danson simulation (/teachermoments/danson)
* Add new indicator for difficult conversation between teacher & parent.
* Add questions for Lori Danson simulation.
* Add summary page that shows all responses at the end of all the scenarios.

* Add logic for hiding timer in audio response.
* Refactor code to ensure the response prompt can be customized in the experience page.
* Fix demo_page and unit tests that were previously failing after adding logic to customize the response prompt.

Bug fix: Adjust min width of responsive_frame to 450 to ensure it does not show a black background around screen.

#### Welcome page
screenshot 2016-12-06 15 49 09

#### One of the scenarios
screenshot 2016-12-06 15 49 29

#### Summary page after responding to all scenarios
screenshot 2016-12-06 15 50 10

### DIFF
--- a/ui/src/components/responsive_frame.jsx
+++ b/ui/src/components/responsive_frame.jsx
@@ -59,6 +59,7 @@ const styles = {
     height: 667,
     overflowY: 'scroll',
     background: 'white',
-    border: '1px solid #999'
+    border: '1px solid #999',
+    overflow: 'auto'
   }
 };
--- a/ui/src/components/responsive_frame.jsx
+++ b/ui/src/components/responsive_frame.jsx
@@ -17,7 +17,7 @@ export default React.createClass({
 
   getDefaultProps() {
     return {
-      minWidth: 400,
+      minWidth: 450,
       minHeight: 400
     };
   },
--- a/ui/src/components/responsive_frame.jsx
+++ b/ui/src/components/responsive_frame.jsx
@@ -59,7 +59,6 @@ const styles = {
     height: 667,
     overflowY: 'scroll',
     background: 'white',
-    border: '1px solid #999',
-    overflow: 'auto'
+    border: '1px solid #999'
   }
 };
--- a/ui/src/message_popup/playtest/danson_experience_page.jsx
+++ b/ui/src/message_popup/playtest/danson_experience_page.jsx
@@ -21,7 +21,6 @@ import ResponsiveFrame from '../../components/responsive_frame.jsx';
 import {withStudents} from '../transformations.jsx';
 
 import * as Api from '../../helpers/api.js';
-import FinalSummaryCard from '../final_summary_card.jsx';
 import NavigationAppBar from '../../components/navigation_app_bar.jsx';
 
 import MobileInterface from '../mobile_prototype/mobile_interface.jsx';
@@ -194,7 +193,6 @@ export default React.createClass({
       obj.questionText = audioResponses[i].question.text.length < maxCharPerLine ? audioResponses[i].question.text : audioResponses[i].question.text.substring(0, maxCharPerLine) + ' ...';
       truncatedAudioResponses.push(obj);
     }
-    console.log(audioResponses);
     return (
       <div className="done">
         <VelocityTransitionGroup enter={{animation: "slideDown"}} leave={{animation: "slideUp"}} runOnMount={true}>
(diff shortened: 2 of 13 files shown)


kesiena115:
@kevinrobinson please take a look

ui/src/message_popup/playtest/danson_experience_page.jsx
```diff
@@ -0,0 +1,344 @@
...
+    return {
+      scaffolding: {
+        helpType: isSolutionMode ? 'none' : 'feedback',
+        shouldShowSummary: !isSolutionMode,
+        hideTimer: true,
+      },
+      gameSession: null,
+      toastRevision: false,
+      limitMs: 90000,
+      email: this.context.auth.userProfile.email
```
kevinrobinson: This changed since the last playtest for some public demos - it's set to `unknown@mit.edu` by default (in the `AuthContainer` component).

You might want to update this for the playtest so that `email` is an empty string if `this.context.auth.userProfile.email` matches `unknown@mit.edu` to make sure the email addresses that are collected are valid.  We can talk out other ways to handle it tomorrow if you want too, it doesn't have to hold up this pull request.

ui/src/message_popup/playtest/danson_experience_page.jsx
```diff
@@ -0,0 +1,344 @@
...
+          <div style={styles.doneTitle}>
+            <p style={styles.paragraph}>You've finished the simulation.</p>
+            <p style={styles.paragraph}><strong>Do not close this page</strong>. You will need it for the reflection.</p>
+            <p style={styles.paragraph}>Please return to the form for the post-simulation reflection.</p>
+          </div>
+          <p style={styles.summaryTitle}>Summary</p>
+          <Divider />
+          {truncatedAudioResponses.map((obj, i) =>
+            <div key={i} style ={_.merge(styles.instructions)}>
+              <p style={styles.paragraph}>{obj.questionText}</p>
```
kevinrobinson: I'm not seeing how the ellipsis comes out here, can you explain?

ui/src/message_popup/playtest/danson_experience_page.jsx
```diff
@@ -0,0 +1,344 @@
+/* @flow weak */
```
kevinrobinson: It might be good to add a single test for this new experience page component.  That would catch things like if we change the API to `PopupQuestion` (like we took advantage of today).  There's a model [here](https://github.com/mit-teaching-systems-lab/threeflows/blob/master/ui/src/message_popup/demo_page_test.jsx) if you have time and want to do that tomorrow, but that doesn't need to hold up this pull request.

ui/src/components/responsive_frame.jsx
```diff
@@ -59,6 +59,7 @@ const styles = {
     width: 375,
     height: 667,
     background: 'white',
-    border: '1px solid #999'
+    border: '1px solid #999',
+    overflow: 'auto'
```
kevinrobinson: Will this conflict with https://github.com/mit-teaching-systems-lab/threeflows/pull/174/files#diff-7d9679f22d80fc6e15010306dd56b8c5L58?

ui/src/message_popup/playtest/danson_experience_page.jsx
kesiena115: It comes out in line 193. The truncatedAudioResponses array exists solely for the purpose of adding ellipsis. It seems like a lot of code for a simple functionality. But that was my quickest hack.

ui/src/message_popup/playtest/danson_experience_page.jsx
kesiena115: I've added a quick fix for this.

ui/src/message_popup/playtest/danson_experience_page.jsx
kesiena115: I'll try to do this tomorrow.

ui/src/message_popup/playtest/danson_experience_page.jsx
kevinrobinson: Ah got it, thanks!  This seems good, or you can do this with [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow) as well if you like.

One thought is that if add the "more" option from your sketch, then the onClick handler can add/remove the CSS style.

PR #176, opened 2016-12-07
Use blobURL (instead of downloadURL) to show the audio in the summary page

kesiena115:
This ensures the user does not have to log in, the playback UI works smoother (i.e. the progress bar moves in sync with the audio), and there is no download icon in the summary page.

## New summary page
screenshot 2016-12-07 13 50 56

#### Before authenticating
screenshot 2016-12-07 12 25 10

#### After authenticating
screenshot 2016-12-06 15 50 10

### DIFF
--- a/ui/src/components/audio_recorder_flow.jsx
+++ b/ui/src/components/audio_recorder_flow.jsx
@@ -93,8 +93,8 @@ export default React.createClass({
   },
 
   onDoneCapture(blob:Blob) {
-    this.props.onLogMessage('message_popup_audio_on_done_capture');
     var downloadUrl = URL.createObjectURL(blob);
+    this.props.onLogMessage('message_popup_audio_on_done_capture', {'blobUrl': downloadUrl});
     this.setState({blob, downloadUrl});
   },
 
--- a/ui/src/message_popup/playtest/danson_experience_page.jsx
+++ b/ui/src/message_popup/playtest/danson_experience_page.jsx
@@ -89,7 +89,7 @@ export default React.createClass({
   },
   
   onLog(type, response:ResponseT) {
-    if (type === 'message_popup_audio_response') {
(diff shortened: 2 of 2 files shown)


ui/src/message_popup/playtest/danson_experience_page.jsx
```diff
@@ -186,12 +186,19 @@ export default React.createClass({
...
+      obj.audioUrl = audioResponses[i].blobUrl;
       obj.questionText = audioResponses[i].question.text.length < maxCharPerLine ? audioResponses[i].question.text : audioResponses[i].question.text.substring(0, maxCharPerLine) + ' ...';
-      truncatedAudioResponses.push(obj);
+      var questionId = audioResponses[i].question.id;
+      if(questionId in reviewedQuestions) {
+        truncatedAudioResponses[reviewedQuestions[questionId]] = obj;
+      } else {
+        reviewedQuestions[questionId] = i;
+        truncatedAudioResponses.push(obj);
+      }
```
kevinrobinson: This probably makes sense as a separate method.

ui/src/components/audio_recorder_flow.jsx
```diff
@@ -93,8 +93,8 @@ export default React.createClass({
   },
 
   onDoneCapture(blob:Blob) {
-    this.props.onLogMessage('message_popup_audio_on_done_capture');
     var downloadUrl = URL.createObjectURL(blob);
+    this.props.onLogMessage('message_popup_audio_on_done_capture', {'blobUrl': downloadUrl});
```
kevinrobinson: This is a great solution for the playtest!  We should revisit this later though, since `onLogMessage` is about logging messages to the server, and these `blobUrl`s will have no meaning outside of this particular browser.

PR #178, opened 2016-12-20
Workaround regression in responsive layout on desktop

kevinrobinson:
Before:
screen shot 2016-12-20 at 3 22 30 pm

After:
screen shot 2016-12-20 at 3 22 24 pm


kevinrobinson:
selfie

PR #179, opened 2017-01-05
Make the scenario prompts expandable and collapsable in the summary page

kesiena115:
There's a default (configurable) char count of 140. The scenario prompt can be displayed in 3 possible ways on the summary page:
* If the scenario prompt is <= 140 chars we display the full text.
* If the char count is > 140, by default the text will be truncated and "... read more" will be appended. The "read more" link can be selected to expand the text.
* If the char count is > 140 and the truncated text gets expanded, the "read more" link changes to a "show less" link which can be selected to truncate the text. 

### Screenshot
screenshot 2017-01-05 13 30 32

### DIFF
--- a/ui/src/message_popup/renderers/audio_response_summary.jsx
+++ b/ui/src/message_popup/renderers/audio_response_summary.jsx
@@ -17,17 +17,19 @@ export default React.createClass({
   },
 
   render() {
+
     const audioResponses = [];
-    const reviewedQuestions = {};  // Keeps track of reviewedQuestions to ensure we only show the last response if the user does multiple retries.
-    this.props.audioResponses.map((arObj, i) => {
+    var lastQuestionId = null;  // For detecting questions with more than one response
+    this.props.audioResponses.map((arObj) => {
       var obj = {};
-      obj.audioUrl = arObj.blobUrl;
-      obj.questionText = arObj.question.text;
       var questionId = arObj.question.id;
-      if(questionId in reviewedQuestions) {
-        audioResponses[reviewedQuestions[questionId]] = obj;
+      if(questionId === lastQuestionId) {
+        var lastObj = audioResponses[audioResponses.length - 1];
+        lastObj.audioUrls.push(arObj.blobUrl);
       } else {
-        reviewedQuestions[questionId] = audioResponses.length;
+        lastQuestionId = questionId;
+        obj.audioUrls = [arObj.blobUrl];
+        obj.questionText = arObj.question.text;
         audioResponses.push(obj);
       }
     });
@@ -45,7 +47,9 @@ export default React.createClass({
           {audioResponses.map((obj, i) =>
             <div key={i} style ={_.merge(styles.instructions)}>
               <ReadMore fulltext={obj.questionText} />
-              <audio controls={true} src={obj.audioUrl} />
+              {obj.audioUrls.map((audioUrl, i) => 
+                <audio key={'audio-url-' + i} controls={true} src={audioUrl} />
+              )}
               <Divider />
             </div>
           )}
--- a/ui/src/message_popup/renderers/audio_response_summary.jsx
+++ b/ui/src/message_popup/renderers/audio_response_summary.jsx
@@ -20,17 +20,17 @@ export default React.createClass({
 
     const audioResponses = [];
     var lastQuestionId = null;  // For detecting questions with more than one response
-    this.props.audioResponses.map((arObj) => {
-      var obj = {};
-      var questionId = arObj.question.id;
+    this.props.audioResponses.map((originalAudioResponse) => {
+      var audioResponse = {};
+      var questionId = originalAudioResponse.question.id;
       if(questionId === lastQuestionId) {
         var lastObj = audioResponses[audioResponses.length - 1];
-        lastObj.audioUrls.push(arObj.blobUrl);
+        lastObj.audioUrls.push(originalAudioResponse.blobUrl);
       } else {
         lastQuestionId = questionId;
-        obj.audioUrls = [arObj.blobUrl];
-        obj.questionText = arObj.question.text;
-        audioResponses.push(obj);
+        audioResponse.audioUrls = [originalAudioResponse.blobUrl];
+        audioResponse.questionText = originalAudioResponse.question.text;
+        audioResponses.push(audioResponse);
       }
     });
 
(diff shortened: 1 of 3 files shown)


kesiena115:
@kevinrobinson  please take a look.

ui/src/message_popup/renderers/audio_response_summary.jsx
```diff
@@ -0,0 +1,94 @@
...
+import ReadMore from './read_more.jsx';
+
+/*
+Component that displays the summary of audio responses.
+*/
+export default React.createClass({
+  displayName: 'AudioResponseSummary',
+
+  propTypes: {
+    gameSession: React.PropTypes.object.isRequired,
```
kevinrobinson: Since `gameSession` has a lot of other things, and this component looks like it only needs the `audioResponses` list, and s it might help simplify to just pass that directly.

ui/src/message_popup/renderers/audio_response_summary.jsx
```diff
@@ -0,0 +1,94 @@
...
+  },
+
+  render() {
+    const audioResponses = [];
+    const reviewedQuestions = {};  // Keeps track of reviewedQuestions to ensure we only show the last response if the user does multiple retries.
```
kevinrobinson: You might be able to express this more directly using `_.uniq` to remove duplicates by question id, or is there something else happening here?

ui/src/message_popup/renderers/audio_response_summary.jsx
```diff
@@ -0,0 +1,94 @@
...
+        <VelocityTransitionGroup enter={{animation: "slideDown"}} leave={{animation: "slideUp"}} runOnMount={true}>
+          <div style={styles.doneTitle}>
+            <p style={styles.paragraph}>You've finished the simulation.</p>
+            <p style={styles.paragraph}><strong>Do not close this page</strong>. You will need it for the reflection.</p>
+            <p style={styles.paragraph}>Please return to the form for the post-simulation reflection.</p>
+          </div>
+          <p style={styles.summaryTitle}>Summary</p>
+          <Divider />
+          {audioResponses.map((obj, i) =>
+            <div key={i} style ={_.merge(styles.instructions)}>
```
kevinrobinson: A better key here is the `questionId`.  Also can you please use a real variable name like `audioResponse` in place of `obj`?

ui/src/message_popup/renderers/read_more.jsx
```diff
@@ -0,0 +1,64 @@
+/* @flow weak */
+import React from 'react';
+
+/*
+  Component that displays a text block that can be expanded or collapsed.
+*/
+export default React.createClass({
```
kevinrobinson: This is a nice way to factor out the little bit of transient state that is needed to track expanded/collapsed! 👍

ui/src/message_popup/renderers/read_more.jsx
```diff
@@ -0,0 +1,64 @@
...
+  },
+
+  collapse() {
+    this.setState({
+      expanded: false
+    });
+  },
+
+  render() {
+    var charCount = (this.props.charCount === undefined) ? 140 : this.props.charCount;
```
kevinrobinson: You can use the React method `getDefaultProps` to express default values like this.

ui/src/message_popup/renderers/read_more.jsx
```diff
@@ -0,0 +1,64 @@
...
+          <a href='#' onClick={this.expand}>read more</a>
+        </div>);
+    }
+  },
+});
+
+const styles = {
+  summaryQuestion: {
+    whiteSpace: 'pre-wrap',
+    fontSize: 16,
```
kevinrobinson: Do you need to set these font styling details here, or could the component that contains this one set these particular values?

ui/src/message_popup/renderers/audio_response_summary.jsx
kesiena115: Fixed

ui/src/message_popup/renderers/audio_response_summary.jsx
kesiena115: Yes there's something more happening. I was trying to ensure we show only the last response. I've actually modified this to show all audio responses for cases where the user does more than one audio recording.

ui/src/message_popup/renderers/audio_response_summary.jsx
kesiena115: I've renamed the variables and changed that key

ui/src/message_popup/renderers/read_more.jsx
kesiena115: Done.

ui/src/message_popup/renderers/read_more.jsx
kesiena115: Oh thanks for that. That was an oversight. I've fixed it.

ui/src/message_popup/renderers/audio_response_summary.jsx
```diff
@@ -13,22 +13,24 @@ export default React.createClass({
...
     const audioResponses = [];
-    const reviewedQuestions = {};  // Keeps track of reviewedQuestions to ensure we only show the last response if the user does multiple retries.
-    this.props.gameSession.audioResponses.map((arObj, i) => {
-      var obj = {};
-      obj.audioUrl = arObj.blobUrl;
-      obj.questionText = arObj.question.text;
-      var questionId = arObj.question.id;
-      if(questionId in reviewedQuestions) {
-        audioResponses[reviewedQuestions[questionId]] = obj;
+    var lastQuestionId = null;  // For detecting questions with more than one response
```
kevinrobinson: Thanks for explaining!  For keeping the last one I think that'd be the same as adding in some `reverse`s with `_.uniq`.

For the change here, you could considering using other higher-level functions to express the intent more directly:
```
responsesByQuestion = _.groupBy(audioResponses, ...);
questionsWithResponses = _.map(responsesByQuestion, ...);
```

Or you could split this code out into a method and name it something that indicates what it will return.  This would more directly express that the shape of the data is changing - here this chunk of code yields the same variable name `audioResponses`, even though that value has a different shape of data than the original `this.props.audioResponses`.

ui/src/message_popup/renderers/audio_response_summary.jsx
```diff
@@ -42,10 +44,12 @@ export default React.createClass({
           </div>
           <p style={styles.summaryTitle}>Summary</p>
           <Divider />
-          {audioResponses.map((obj, i) =>
-            <div key={i} style ={_.merge(styles.instructions)}>
-              <ReadMore fulltext={obj.questionText} />
-              <audio controls={true} src={obj.audioUrl} />
+          {audioResponses.map((audioResponse) =>
+            <div key={audioResponse.questionId} style ={_.merge(styles.instructions, styles.summaryQuestion)}>
```
kevinrobinson: You can also use destructuring to merge to objects like `style={{...styles.instructions, ...styles.summaryQuestion}}` in place of `_.merge`.  That's more in the style of JSX and ES6 and I hadn't shifted over in older code.

kevinrobinson:
Also, since sometimes GitHub is confusing, this is approved!

screen shot 2017-01-06 at 8 19 39 am

So while I've left you some suggestions it's up to you how to proceed, and however you decide to address them with comments or changes, you can merge when you feel ready. 👍

PR #182, opened 2017-01-12
Fix Flow bug with importing type along with code

kevinrobinson:
Fix for https://travis-ci.org/mit-teaching-systems-lab/threeflows/builds/191430161


kevinrobinson:
selfie

PR #184, opened 2017-01-17
Add text fallback for MinimalAudioResponse when audio not supported

kevinrobinson:
This affects mobile and desktop Safari.  Without feature detection, tapping the "Record" button is a JS error and halts execution.

Fallback to text:
screen shot 2017-01-17 at 10 53 54 am


kevinrobinson:
selfie

PR #186, opened 2017-01-18
Add MindsetKit scenario, factor out MinimalOpenResponse

kevinrobinson:
A single-shot scenario, adapted from https://www.mindsetkit.org/belonging/cues-belonging/quiz-classroom-scenarios

screen shot 2017-01-18 at 12 42 40 pm


kevinrobinson:
selfie

PR #187, opened 2017-01-20
Add a new endpoint and experiment with a Mechanical Turk HIT

kevinrobinson:
This is an experiment to add the plumbing necessary to run this as a Mechanical Turk HIT.  We aren't running any actual experiments and this is only intended to learn how it would work and try things out in the MTurk Requester Sandbox.

This creates a new endpoint, and copies parts of `InsubordinationPage` out into `InsubordinationExperiment`, which removes bits related to requesting the user's email and nav bar for restarting the session.  A new `MTurkPage` component takes params from Mechanical Turk, passes an opaque `userIdentifier` along to the experiment page along with an `onLogMessage` function that mixes in all the Mechanical Turk information when logging.  Finally, an `onExperimentDone` page allows the experiment to signify it's done, and `MTurkPage` completes the call to Amazon to finish the HIT.

Screenshot, with frame removed:
screen shot 2017-01-20 at 12 54 53 pm

### DIFF
--- a/ui/src/app.jsx
+++ b/ui/src/app.jsx
@@ -21,7 +21,6 @@ import * as MessagePopup from './message_popup/index.js';
 import {challenges, slates} from './challenge/challenges.js';
 
 
-
 export default React.createClass({
   displayName: 'App',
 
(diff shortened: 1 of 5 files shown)


kevinrobinson:
selfie, these changes are isolated to a new route and needs to be deployed to test in the sandbox

PR #188, opened 2017-01-20
Revise MTurkPage to support preview mode

kevinrobinson:
Per "Preview mode" in http://docs.aws.amazon.com/AWSMechTurk/latest/AWSMturkAPI/ApiReference_ExternalQuestionArticle.html, the params passed are different in preview mode before the HIT is accepted.

 screen shot 2017-01-20 at 2 39 22 pm


kevinrobinson:
selfie

PR #189, opened 2017-01-23
Fix production build of JS

kevinrobinson:
This wasn't actually building the production mode, so fix that and minify and strip out `process.env` checks.


kevinrobinson:
selfie

PR #190, opened 2017-01-23
Need uglify and envify for production build

kevinrobinson:
Follow-on from https://github.com/mit-teaching-systems-lab/threeflows/pull/189


kevinrobinson:
selfie

PR #192, opened 2017-02-08
Fix background on root endpoint, fix image URL bug in Twine demo

kevinrobinson:
## Before:
screen shot 2017-02-07 at 11 31 19 pm

## After:
screen shot 2017-02-07 at 11 31 22 pm

screen shot 2017-02-07 at 11 37 46 pm


kevinrobinson:
selfie

PR #193, opened 2017-02-16
Add more logging around S3 being unavailable

kevinrobinson:
The server now tolerates S3 being unavailable when reading audio files (eg., when no internet connectivity), and returns a 503.

The client code for recording audio logs an error to the console if the POST request to the server fails (eg., when no internet connectivity).

Making the client more robust to failures involves other changes (since that would mean actual data loss); for now this just logs more explicitly.


kevinrobinson:
selfie

PR #194, opened 2017-02-16
Move positioning of 'Recording...' text so buttons don't jump

kevinrobinson:
This moves the text below the button, since in the case where there's no prompt text, the button would jump when "Recording..." was added during the state transition.

After:

screen shot 2017-02-16 at 8 49 12 am


kevinrobinson:
selfie

PR #195, opened 2017-02-16
Back out state change for upload error, just log

kevinrobinson:
Soften https://github.com/mit-teaching-systems-lab/threeflows/pull/193 to just logging, since there are nice Flow states that correctly show doing this would stop the UI flow from proceeding (which we'll accept for now).


kevinrobinson:
selfie

PR #197, opened 2017-02-16
Update Rollbar

kevinrobinson:
And try swapping instance/class names.  I can't tell what's going on but there's a different between production and development, and it appears unable to find methods on Rollbar.  Changing this to exactly match the docs.


kevinrobinson:
selfie

PR #198, opened 2017-02-16
Add Branching to Lori Danson Simulation

kesiena115:
* The new endpoint is teachermoments/danson2
* Create new LinearSession for Lori Danson simulation (mostly a copy of the regular LinearSession) with minor changes.
* Refactor code to include prop for forcing audio response in places where it was missing
* Refactor code to add selected choice in the submitted response

### DIFF
--- a/ui/src/message_popup/linear_session/danson_linear_session.jsx
+++ b/ui/src/message_popup/linear_session/danson_linear_session.jsx
@@ -1,0 +1,83 @@
+/* @flow weak */
+import React from 'react';
+import VelocityTransitionGroup from "velocity-react/velocity-transition-group";
+import 'velocity-animate/velocity.ui';
+
+
+// Manages the flow through a list of questions.
+// Starts with `introEl`, then moves through showing `questionEl` for each
+// question and collecting a response and logging it, then showing
+// `summaryEl` when done.
+export default React.createClass({
+  displayName: 'DansonLinearSession',
+
+  propTypes: {
+    questions: React.PropTypes.array.isRequired,
+    questionEl: React.PropTypes.func.isRequired,
+    summaryEl: React.PropTypes.func.isRequired,
+    onLogMessage: React.PropTypes.func.isRequired
+  },
+
+  getInitialState() {
+    return {
+      responses: []
+    };
+  },
+
+  onResetSession() {
+    this.setState(this.getInitialState());
+  },
+
+  // Mixes in question to payload
+  onLogMessage(question, type, rawResponse) {
+    this.props.onLogMessage(type, {...rawResponse, question});
+  },
+
+  // Logs and transitions
+  onResponseSubmitted(question, response) {
+    this.onLogMessage('on_response_submitted', response);
+    this.setState({
+      responses: this.state.responses.concat(response)
+    });
+  },
+
+  // Animation wrapper
+  render() {
+    return (
+      <VelocityTransitionGroup
+        leave={{animation: "slideUp", duration: 200}}
+        enter={{animation: "slideDown", duration: 200}}
+        runOnMount={true}
+      >
+        {this.renderContent()}
+      </VelocityTransitionGroup>
+    );
+  },
+
+  renderContent() {
+    const {questions} = this.props;
+    const {responses} = this.state;
+    if (responses.length >= questions.length) return this.props.summaryEl(questions, responses);
+
+    if(responses.length == 3) {
+      const lastQuestionIndex = responses.length - 1;
+      const allChoices = questions[lastQuestionIndex].choices;
+      const selectedChoice = responses[responses.length - 1].choice;
+      if(selectedChoice === allChoices[0] || selectedChoice === allChoices[1]) {
+        this.props.questions.splice(lastQuestionIndex + 1, 1);
+      } else if(selectedChoice === allChoices[2]) {
+        this.props.questions.splice(lastQuestionIndex + 2, 1);
+      } else {
+        this.props.questions.splice(lastQuestionIndex + 1, 2);
+      }
+    }
+
+    const question = questions[responses.length];
+    return (
+      <div key={question.id}>
+        {this.props.questionEl(question, this.onLogMessage.bind(this, question), this.onResponseSubmitted.bind(this, question))}
+      </div>
+    );
+  }
+});
+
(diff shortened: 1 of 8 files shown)


ui/src/message_popup/playtest/danson_experience_page2.jsx
```diff
@@ -0,0 +1,181 @@
...
+
+    return {
+      email,
+      questions: null,
+      sessionId: uuid.v4()
+    };
+  },
+
+  onStart(email) {
+    const questions = InsubordinationScenarios.questions();
```
kevinrobinson: Copy-paste bug on the name here.

ui/src/message_popup/playtest/danson_experience_page2.jsx
```diff
@@ -0,0 +1,181 @@
...
+            <p>Welcome!</p>
+            <p>You will go through a set of scenarios that simulate the conversation between you and Mrs. Danson.</p>
+            <p>You need to provide an audio response to each prompt. Please respond as quickly as possible (like you would do in a real conversation).</p>
+            <p>This may feel uncomfortable at first, but it's better to feel uncomfortable here than with a real parent.</p>
+        </div>
+      </IntroWithEmail>
+    );
+  },
+
+  // Only ask for audio on questions with choices, otherwise let them continue
```
kevinrobinson: Stale comment.

ui/src/message_popup/playtest/danson_scenarios.js
```diff
@@ -0,0 +1,56 @@
...
+
+export type QuestionT = {
+  id:number,
+  choices:[string],
+  text:string
+};
+
+
+// Make questions and choices.
+export const InsubordinationScenarios = {
```
kevinrobinson: Copy-paste leftover on the name.

ui/src/message_popup/playtest/danson_experience_page2.jsx
```diff
@@ -0,0 +1,181 @@
...
+            <p style={styles.paragraph}><strong>Do not close this page</strong>. You will need it for the reflection.</p>
+            <p style={styles.paragraph}>Please return to the form for the post-simulation reflection.</p>
+          </div>
+          <p style={styles.summaryTitle}>Summary</p>
+          <Divider />
+
+          {responses.map((response, i) => 
+            <div key={"question-" + i} style ={_.merge(styles.instructions, styles.summaryQuestion)}>
+              <ReadMore fulltext={questions[i].text}/>
+              <audio key={'response-' + i} controls={true} src={response.audioResponse.downloadUrl} style={{paddingTop: 10, paddingBottom: 20}}/>
```
kevinrobinson: I'm not entirely following the change of adding `downloadUrl` everywhere.  I was expecting that you could just look at `uploadedUrl` here and use that, or is it important to thread through the local `downloadUrl`, maybe the time to load the uploaded URL from the server matters?

ui/src/message_popup/playtest/danson_experience_page2.jsx
```diff
@@ -0,0 +1,181 @@
...
+        <VelocityTransitionGroup enter={{animation: "slideDown"}} leave={{animation: "slideUp"}} runOnMount={true}>
+          <div style={styles.doneTitle}>
+            <p style={styles.paragraph}>You've finished the simulation.</p>
+            <p style={styles.paragraph}><strong>Do not close this page</strong>. You will need it for the reflection.</p>
+            <p style={styles.paragraph}>Please return to the form for the post-simulation reflection.</p>
+          </div>
+          <p style={styles.summaryTitle}>Summary</p>
+          <Divider />
+
+          {responses.map((response, i) =>
```
kevinrobinson: This might be a good place to look at refactoring `AudioResponseSummary` after to split out the code that shows the summary for a list of question texts and audio URLs.

ui/src/message_popup/linear_session/record_then_classify_question.jsx
```diff
@@ -29,11 +30,27 @@ export default React.createClass({
   },
 
   onDoneAudio(audioResponse) {
-    this.setState({audioResponse});
+    if(this.props.question.choices.length > 0) {
```
kevinrobinson: If there are no choices, this will change the behavior the component in a way that will affect the user experience in other pages, which I don't think we want.

Could you instead branch in `DansonExperiencePage2#renderQuestionEl` and check if there are choices?  If there are choices, use `RecordThenClassify` as it is now without any changes.  And if there are choices, use `MinimalOpenResponse` directly.  Would something like that help achieve what we want, without changing this component's behavior and affecting other pages?

Something like this is how [InsubordinationPage#renderQuestionEl](https://github.com/kesiena115/threeflows/blob/ddeaba5f45b122401ff8f65fdba6425b7c5382af/ui/src/message_popup/playtest/insubordination_page.jsx#L109) works.

ui/src/message_popup/linear_session/danson_linear_session.jsx
```diff
@@ -0,0 +1,83 @@
...
+    const {questions} = this.props;
+    const {responses} = this.state;
+    if (responses.length >= questions.length) return this.props.summaryEl(questions, responses);
+
+    if(responses.length == 3) {
+      const lastQuestionIndex = responses.length - 1;
+      const allChoices = questions[lastQuestionIndex].choices;
+      const selectedChoice = responses[responses.length - 1].choice;
+      if(selectedChoice === allChoices[0] || selectedChoice === allChoices[1]) {
+        this.props.questions.splice(lastQuestionIndex + 1, 1);
```
kevinrobinson: Mutating `props` is never a good idea in React - it needs to be able to track changes to `props` and `state` explicitly in order to update the DOM.  If you mutate those values directly, the DOM may not be in sync with what you expect based on passing props and setting state.

Here, something that might work well is "there's some function that takes (`responses`, `questions`) and then returns `question`.  You could factor this code out into a method, and then return the value for `question` without having to mutate the props.

After the playtest, we could pull that out into a prop that `LinearSession` takes, and then we can remove the duplication too.

ui/src/message_popup/playtest/danson_experience_page2.jsx
kesiena115: The problem with loading from the server is that it'd request for the username and password (that's what I noticed last time). Also, the playback tends to not be smooth i.e. it was a bit challenging to pause/skip to exact timestamps.

kevinrobinson: Got it, thanks for explaining!  👍

PR #199, opened 2017-02-16
Fix message type logging bug for forced choice and onSubmit in LinearSession

kevinrobinson:
This was logging all the data, but incorrectly serializing some response data into the message type.


kevinrobinson:
selfie, cc @kesiena115

PR #200, opened 2017-02-17
Add more lesson plan context to CS sub

kevinrobinson:
So this would make sense in a self-serve way, without screen shot 2017-02-17 at 10 54 06 am

Also remove multiline text on email sign-up.


kevinrobinson:
selfie

PR #203, opened 2017-03-01
Remove older faculty pages and related server endpoints

kevinrobinson:
We aren't actively using these right now, and removing them makes for stricter data access controls.


kevinrobinson:
selfie

PR #205, opened 2017-03-01
Restore uploaded URL for audio as an identifier

kevinrobinson:
Follow on from https://github.com/mit-teaching-systems-lab/threeflows/pull/203; client code expects `url` to be in the response.  We should change that to just use identifiers, but restoring it for now to tackle separately.


kevinrobinson:
selfie

PR #208, opened 2017-03-01
Switch to Uglify2 to address Rollbar mangling

kevinrobinson:
Rollbar isn't working in production.  It works locally, and the code is failing within Rollbar's code while trying to do a property lookup, so the problems appears to be introduced in the production build process.

In looking at that, I realized this was using Uglify so updated it to Uglify2, and that appears to resolve the problem in a production build locally.


kevinrobinson:
selfie

PR #209, opened 2017-03-01
Bump React and swap to uglifyify

kevinrobinson:
Follow on from https://github.com/mit-teaching-systems-lab/threeflows/pull/208.


kevinrobinson:
selfie

PR #210, opened 2017-03-01
Move Uglify to dependencies

kevinrobinson:
One more for https://github.com/mit-teaching-systems-lab/threeflows/pull/209, moving this into dependencies so it's installed in a Heroku deploy and available to then build the UI code.


kevinrobinson:
selfie

PR #213, opened 2017-03-02
Add guard for AudioRecorder#stop

kevinrobinson:
Addressing https://rollbar.com/threeflows/threeflows/items/97/occurrences/22166214916/  There's a lot of state transitions here, it'd be better to encode these more explicitly and potentially remove some (and clarify if `destroy` is needed or whether `record/stop` can encode all the states from the caller's perspective).  For now just addressing the immediate issue.


kevinrobinson:
Closing, this got stale.

PR #218, opened 2017-03-09
Add in Common Beliefs Survey to discipline

kevinrobinson:
From SPLC's [Teaching Tolerance](http://www.tolerance.org/sites/default/files/general/Common%20Beliefs%20Survey%20New_0.pdf)

Example:
screen shot 2017-03-09 at 2 33 00 pm


kevinrobinson:
selfie

PR #221, opened 2017-03-21
Revisions for Thursday class copy

kevinrobinson:
This changes some of the context around the scenario for a class Thursday.  It's accessible at `/teachermoments/thursday`.  It also adds URL and client timestamp logging in `Api.logEvidence`.

# At the beginning
screen shot 2017-03-21 at 4 46 02 pm
screen shot 2017-03-21 at 4 46 30 pm

## At the end
screen shot 2017-03-21 at 4 49 21 pm
screen shot 2017-03-21 at 4 50 04 pm


kevinrobinson:
selfie

PR #225, opened 2017-03-23
Refactor linear_session to work with Lori Danson simulation by including a function for getting the next question

kesiena115:


ui/src/message_popup/linear_session/linear_session.jsx
```diff
@@ -57,9 +69,13 @@ export default React.createClass({
   renderContent() {
     const {questions} = this.props;
     const {responses} = this.state;
-    if (responses.length >= questions.length) return this.props.summaryEl(questions, responses);
 
-    const question = questions[responses.length];
+    if (this.props.getNextQuestion === undefined || this.props.getNextQuestion(questions, responses) === null) {
```
kevinrobinson: With `getDeaultProps`, I don't think you need to do this check for whether it's undefined.   Also, the way this is written now you could call `getNextQuestion` twice: one here and once on line 77.

I think it'd be clearer using something like `const question = this.props.getNextQuestion(...)` and then checking that variable's value.

ui/src/message_popup/linear_session/linear_session.jsx
```diff
@@ -14,10 +14,22 @@ export default React.createClass({
   propTypes: {
     questions: React.PropTypes.array.isRequired,
     questionEl: React.PropTypes.func.isRequired,
+    getNextQuestion: React.PropTypes.func,
```
kevinrobinson: This is a bit of a nitpick, but using `onQuestionNeeded` might be a event-y sounding name to signify that this is callback function.

kevinrobinson: oops, scratch that, this is called in `render`

ui/src/message_popup/playtest/danson_experience_page.jsx
```diff
@@ -69,6 +69,32 @@ export default React.createClass({
     });
   },
 
+  onGetNextQuestion(questions:[QuestionT], responses:[ResponseT]) {
```
kevinrobinson: This looks great, but could use a one line comment at the top of the method about why there's this logic here.

ui/src/message_popup/playtest/danson_experience_page.jsx
```diff
@@ -69,6 +69,32 @@ export default React.createClass({
...
+    const allQuestions = questions.slice();
```
kevinrobinson: 👍  Nice, this is great so you can work with `questions` but not modify the value passed into this function.

ui/src/message_popup/linear_session/linear_session.jsx
```diff
@@ -14,10 +14,22 @@ export default React.createClass({
...
     summaryEl: React.PropTypes.func.isRequired,
     onLogMessage: React.PropTypes.func.isRequired
   },
 
+  getDefaultProps() {
```
kevinrobinson: This is a small style suggestion, but if a "default method" gets more involved you can also split it out to be a method on the component and call it here, or a pure function at the top of the file.

ui/src/message_popup/linear_session/linear_session.jsx
kesiena115: Even with `getDefaultProps`, flow still throws an error if I don't check that the function is not undefined as shown in the screenshot below. I only added the check for getNextQuestion in order to eliminate that error from flow. That error actually causes the build to fail the automated tests run on github. I've found a [github issue](https://github.com/facebook/flow/issues/1660) related to this but don't yet know how to work around it without the explicit check. Perhaps we can look at this together

screenshot 2017-03-30 19 38 52

kevinrobinson: Ah got it, thanks for explaining!  Yeah seems like an edge case bug between React and Flow, where Flow can't tell React will guarantee that `this.props.getNextQuestion` will have a value by that point.

For these "possibly undefined values" warnings you can explicitly check that it's defined with `const question = this.props.getNextQuestion && this.props.getNextQuestion(...)`.

For those other warnings, looks like just a `handleLayout: ` with a colon instead of an equals should help.  If those don't work then feel free to drop another comment or we can take a look together.

ui/src/message_popup/linear_session/linear_session.jsx
kesiena115: Fixed. I'm not 100% sure I've done it the way you wanted though, so you can take a look and let me know.

ui/src/message_popup/playtest/danson_experience_page.jsx
kesiena115: Done

server/index.js
```diff
@@ -57,7 +57,9 @@ function questionAuthoringAuth(req, res, next) {
 // api routes
 // helper for db connection pooling
 function queryDatabase(text, values, cb) {
-  pg.connect(process.env.DATABASE_URL +'?ssl=true', function(err, client, done) {
+  var addSSL = process.env.NODE_ENV === 'development' ? '' : '?ssl=true';
+  pg.connect(process.env.DATABASE_URL + addSSL, function(err, client, done) {
+    console.log(err)
```
kevinrobinson: I'd either remove this, or wrap it in an `if` so it only logs if there actually is an error.

ui/src/message_popup/linear_session/linear_session.jsx
kevinrobinson: Looks good!

PR #230, opened 2017-04-04
Add limited review flow for learners

kevinrobinson:
#### Summary
Allow learners to review data from their responses.  They must have two access codes that they receive in-person or out of band, and then must authenticate their email address.  Then they're given a token that can be used for a limited time window to review their response data.

#### Start review
They need the URL and the access code from their facilitator:

#### Receive email
An authentication step:

#### Follow link from email and confirm
Confirm their email again.  The (token, hashed email) pair grants access to data:

#### Review data
For use in-person, or individually before a group debrief:

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,8 +99,7 @@ $ heroku addons:create heroku-postgresql:hobby-dev
 
 Seed the database:
 ```
-$ heroku pg:psql
-threeflows:DATABASE=> CREATE TABLE evidence (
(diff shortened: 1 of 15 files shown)


kevinrobinson:
Database has been migrated already.  Images aren't uploading remotely, going to switch locales and update after.

kevinrobinson:
selfie

PR #231, opened 2017-04-04
Patch bug with renaming audio streaming method

kevinrobinson:
Follow on to https://github.com/mit-teaching-systems-lab/threeflows/pull/230


kevinrobinson:
selfie

PR #232, opened 2017-04-04
Add key to audio component to prevent re-rendering and reloading

kevinrobinson:
Left out mistakenly in https://github.com/mit-teaching-systems-lab/threeflows/pull/230.


kevinrobinson:
selfie

PR #233, opened 2017-04-04
Add button to listen to audio, rather than loading all at once

kevinrobinson:
Improvement for https://github.com/mit-teaching-systems-lab/threeflows/pull/230, so that all audio files aren't downloaded in parallel at once before they're needed.  Also updates `autoPlay=true` so that audio starts as soon as it's available after clicking "Listen" button.

Separately, updates the way the token query string is removed from the URL to use `replaceState` instead of `pushState`.


kevinrobinson:
#### Before clicking
screen shot 2017-04-04 at 1 24 03 pm


#### After clicking, with autoPlay
screen shot 2017-04-04 at 1 22 18 pm

kevinrobinson:
selfie

PR #236, opened 2017-04-05
Add global kill switch for reviewing responses

kevinrobinson:
Add `IS_SENSITIVE_REVIEWING_ENABLED` env variable to allow global kill switch disabling any reviewing capabilities.  The database rows can control this at a more granular level, but this is an additional layer since we typically only open reviews for very limited time windows right now.

Also adds some more server logging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   },
   "main": "server/index.js",
   "scripts": {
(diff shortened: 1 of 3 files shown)


kevinrobinson:
selfie

PR #237, opened 2017-04-05
Fix bug with checking reviewTokenRow when kill switch

kevinrobinson:
Fix bug from https://github.com/mit-teaching-systems-lab/threeflows/pull/236.


kevinrobinson:
selfie

PR #238, opened 2017-04-05
Script for grabbing data for offline analysis

kevinrobinson:
Pull out data from the database for analysis offline.  Analysis tools can then be built to use this data format and can be committed to source control, so that this analysis isn't ever available through the web.

The alternative would be to build analysis tools that query the database and deploy it to the web behind an authorization check, but this adds some level of participant risk.  This will let us develop analysis tools quickly, but limit and control more tightly who has access to them.

### DIFF
--- a/scripts/clone.js
+++ b/scripts/clone.js
@@ -1,0 +1,74 @@
+const pg = require('pg');
+
+
+// query via connection pool
+function queryDatabase(text, values, cb) {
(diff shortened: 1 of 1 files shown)


kevinrobinson:
selfie

kevinrobinson:
cc @kesiena115 in case this is useful for you as well.

PR #239, opened 2017-04-06
Fix bug in linear_session, add test to confirm fix

kevinrobinson:
This fixes a bug in https://github.com/mit-teaching-systems-lab/threeflows/pull/225, which affects all experience pages that don't provide `getNextQuestion` (eg., `/sub` endpoint).

@kesiena115 This is my bad, I think I gave you bad advice in the PR, this essentially restores it to the way you had done it originally where if there's no prop a default method is called.  I also added tests verifying this.


kevinrobinson:
selfie, fixing broken state in production

PR #240, opened 2017-04-12
Prevent default for demos button

kevinrobinson:
On mobile, the touch ripple leads to extra taps coming through after the re-rendering.


kevinrobinson:
selfie

PR #242, opened 2017-04-12
Add notice about consent inline, and component for consent inline after scenario

kevinrobinson:
This adds a notice about consent to `IntroWithEmail`, so that this is shown to all participants even if they arrive to the experience self-serve or inbound from an article or elsewhere.  It adds `ResearchConsent` that shows the consent form inline and adds that to the end of the CS substitute scenario.

Finally, it also adds an inline consent form that can be used standalone, if this conflicts with the flow of doing the experience, reflecting and then rejoining the group.  In that case, the check at the end can be an additional reminder.

## Notice shown up front
screen shot 2017-04-12 at 9 47 27 am

## Explanation at the end of the scenario
screen shot 2017-04-12 at 10 43 20 am

## Completing consent form inline
screen shot 2017-04-12 at 10 44 51 am
screen shot 2017-04-12 at 10 44 33 am

## If the user declines
screen shot 2017-04-12 at 10 44 42 am

## Standalone at /consent
screen shot 2017-04-12 at 11 06 37 am

### DIFF
--- a/ui/src/app.jsx
+++ b/ui/src/app.jsx
@@ -10,6 +10,7 @@ import AuthContainer from './auth_container.jsx';
 import VirtualSchoolPage from './virtual_school/virtual_school_page.jsx';
 import HomePage from './home/home_page.jsx';
 import DemosPage from './home/demos_page.jsx';
+import ConsentPage from './home/consent_page.jsx';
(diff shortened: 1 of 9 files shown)


kevinrobinson:
selfie

PR #244, opened 2017-04-12
Tweak consent copy

kevinrobinson:
Also wrap `IntroWithEmail` so there's only one element on the page that animates, not two independent elements.

screen shot 2017-04-12 at 12 30 35 pm


kevinrobinson:
selfie

PR #247, opened 2017-04-21
Add log message timestamp at the instant that each Jennifer Turner scenario finishes playing

kesiena115:


ui/src/message_popup/playtest/turner_experience_page.jsx
```diff
@@ -68,6 +68,13 @@ export default React.createClass({
 
   onScenarioDone() {
     this.setState({recording: true});
+    
+    Api.logEvidence('finished_playing_scenario', {
```
kevinrobinson: Could you call `this.onLogMessage` here instead?   It adds in the sessionId,  timestamp, etc., and then all logging is going through that one central point.

PR #250, opened 2017-04-23
Refactor code and add two new components for Jennifer Turner Scenario

kesiena115:
* Each audio recording begins immediately the video finishes playing i.e. the user does not need  to click any button.
* We show a still image of Jennifer Turner during each audio recording.

## Screenshot of a video scenario
screenshot 2017-04-22 20 20 20

## Still image while recording response
screenshot 2017-04-22 20 20 28

## Summary page
screenshot 2017-04-22 20 22 46

### DIFF
--- a/ui/src/message_popup/renderers/instant_audio_recorder.jsx
+++ b/ui/src/message_popup/renderers/instant_audio_recorder.jsx
@@ -1,0 +1,144 @@
+/* @flow weak */
+import React from 'react';
+import RaisedButton from 'material-ui/RaisedButton';
+import AudioCapture from '../../components/audio_capture.jsx';
+
+type State = {
+  haveRecorded: bool,
+  downloadUrl: ?string,
+  uploadState: 'idle' | 'pending' | 'done',
+  blob: ?Blob,
+  uploadedUrl: ?string
+};
+
+export default React.createClass({
+  displayName: 'InstantAudioRecorder',
+
+  propTypes: {
+    url: React.PropTypes.string.isRequired,
+    onDoneUploading: React.PropTypes.func.isRequired,
+    onLogMessage: React.PropTypes.func.isRequired,
+    instantRecord: React.PropTypes.bool,
+    recordingPrompt: React.PropTypes.string,
+    recordingText: React.PropTypes.string,
+  },
+
+  getDefaultProps() {
+    return {
+      instantRecord: false,
+      recordingPrompt: "Recording...",
+      recordingText: "Done",
+    };
+  },
+
+  getInitialState():State {
+    return {
+      haveRecorded: false,
+      downloadUrl: null,
+      uploadState: 'idle',
+      blob: null,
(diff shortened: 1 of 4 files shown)


ui/src/message_popup/renderers/instant_audio_recorder.jsx
```diff
@@ -0,0 +1,144 @@
...
+
+type State = {
+  haveRecorded: bool,
+  downloadUrl: ?string,
+  uploadState: 'idle' | 'pending' | 'done',
+  blob: ?Blob,
+  uploadedUrl: ?string
+};
+
+export default React.createClass({
```
kevinrobinson: If I read this right, this is the same as [AudioRecorderFlow](https://github.com/mit-teaching-systems-lab/threeflows/blob/master/ui/src/components/audio_recorder_flow.jsx), but with some steps in the process missing, is that right?

It'd be good to call out the difference and compare them in a comment at the top of the class here.  There might be opportunities to generalize, but it seems like the essence of these two components are their transitions between the different states so also fine to keep them separate.

ui/src/message_popup/renderers/instant_audio_recorder.jsx
```diff
@@ -0,0 +1,144 @@
...
+
+    if (!this.props.instantRecord && !haveRecorded) return 'idle';
+    if (this.props.instantRecord && !haveRecorded) return 'recording';
+    if (haveRecorded && !blob) return 'processing';
+    if (blob && uploadState === 'pending') return 'saving';
+    if (blob && uploadedUrl) return 'done';
+  },
+
+  onDoneRecordingClicked() {
+    this.props.onLogMessage('message_popup_audio_done_recording_clicked');
```
kevinrobinson: We should chat after the playtest about the message type strings here - right now if they're the same as other components, so in looking at the data on the server there's no way to distinguish between recording that was done with the instant component or the "review" one.  We might want to use different message types per component to capture more of this, but the tradeoff is more message types when doing analysis.

ui/src/message_popup/renderers/instant_audio_recorder.jsx
```diff
@@ -0,0 +1,144 @@
...
+    this.setState({
+      // isRecording: false,
```
kevinrobinson: Stray comment here.

ui/src/message_popup/renderers/instant_response_scenario.jsx
```diff
@@ -0,0 +1,123 @@
...
+              controls: 0,
+              rel: 0,
+              showinfo: 0,
+              start: question.start,
+              end: question.end
+            }
+          }} 
+        />
+        <div style={styles.invisible}>
+          { /* Todo (Kes): This is a hack to minimize wait time on first image display */ }
```
kevinrobinson: Nice touch! 👍

PR #257, opened 2017-06-06
Add logging to video summary & video scenarios

kesiena115:
Add logging when a video scenario gets paused or played and when either video scenarios or audio responses get played during the post-simulation reflection


kevinrobinson:
@kesiena115 approved, ship it!

PR #261, opened 2017-06-14
Add query param to jump to particular scenes for Danson and Turner

kevinrobinson:
This is for testing and demoing (eg., for CSCL).

Examples:
```
/teachermoments/turner?p=5
/teachermoments/danson?p=6
```


kevinrobinson:
selfie

PR #264, opened 2017-06-19
Add initial Jayden counseling prototype

kevinrobinson:
Adds the initial counseling prototype from KJ's slides.  Also factor out `QuestionInterpreter` for re-use to standardize some of the question formats.

Excerpt:
screen shot 2017-06-19 at 6 29 15 pm


kevinrobinson:
selfie

PR #270, opened 2017-06-23
Add audio and text summary screen to Jayden scenario

kevinrobinson:
The ending of the Jayden scenario is reworked for the playtest next week.

This also makes three other small fixes:
- `QuestionInterpreter` checks audio recording capabilities, and falls back to "Respond" when there's no audio recording.
- `LinearSession` always merges in the question as part of the `responses` it collects.
- query strings are passed along to Jayden and other pages in `app.jsx`

## Changes
screen shot 2017-06-23 at 2 45 21 pm

screen shot 2017-06-23 at 2 32 17 pm

### DIFF
--- a/ui/src/app.jsx
+++ b/ui/src/app.jsx
@@ -125,19 +125,19 @@ export default React.createClass({
   },
 
   alphaPlaytest(query = {}) {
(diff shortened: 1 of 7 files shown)


kevinrobinson:
selfie

PR #271, opened 2017-06-23
Switched Mr. Smith scenario over to bubbleSort

milsim11:
Added a Mr. SmithA which is the same scenario but it in a different location (I wanted to switch it over to use the bubbleSort and question interpreter format). 
 
Also cleaned it up (formatting and grammar changes)

Added grammar and formatting changes to original smith (turner based). 

Also added minimal text response to question interpreter so that scenarios can have text box responses.


ui/src/message_popup/playtest/smithA_scenario.jsx
```diff
@@ -0,0 +1,401 @@
+/* @flow weak */
+import React from 'react';
+
+/*
+This file defines the content for the scenario around bubble sort.
```
kevinrobinson: copy-pasta comment here

ui/src/message_popup/playtest/smithA_scenario.jsx
```diff
@@ -0,0 +1,401 @@
...
+    <div>A little bit of background information on the lesson being taught.</div>
+    <br />
+    <div>3. Try it!</div>
+    <div>When you're ready, you'll go through a set of short scenes that simulate moments in the classroom. Note what you observe.</div>
+    <br />
+    <div>4. PAUSE!</div>
+    <div>Pause for a brief group discussion before resuming.</div>
+    <br />
+    <div>5. Lenses</div>
+    <div>Try another set of scenes, this time from a new perspective.</div>
```
kevinrobinson: Is this accurate, that there's another set of scenes?

ui/src/message_popup/playtest/smithA_scenario.jsx
```diff
@@ -0,0 +1,401 @@
...
+
+In doing so, his students will be working with the standard internet encoding protocol, ASCII. The class will be using pair programming to complete a project developing their own protocol for sending formatted (i.e bold, italicized, etc.) text. 
+
+      `
+    });
+
+  slides.push({type: 'Background', el: 
+<div>
+  <div>Useful Knowledge:</div>
+  <br />
```
kevinrobinson: This scenario uses a lot more `el` scenes, where you're passing React elements.  We can probably figure out a better way to structure that, since when we're writing rendering code we'd usually use styles or CSS to control the spacing rather than adding `<br />` tags like the original example you copied from does (my old mistake).  Let's leave that for now and we can revisit afterward.

PR #272, opened 2017-06-26
Add revised ECS scenario

kevinrobinson:
This revises the CS sub scenario for ECS folks.  It shortens the scenario slightly, revises the anticipate and reflection questions to be more squarely about equity and inclusiveness, and makes a few other small changes like adding student pairs in more places.

It also adds a prop to `MinimalTextResponse` to allow setting the height.

## Screenshots
screen shot 2017-06-26 at 8 26 54 am
screen shot 2017-06-26 at 8 21 26 am


kevinrobinson:
selfie

PR #274, opened 2017-06-26
Revise copy to AudioResponseSummary for Jayden scenario

kevinrobinson:
Also make summary text a prop for `AudioResponseSummary`, since this is scenario-specific.


kevinrobinson:
selfie

PR #276, opened 2017-06-27
Allow forcing text responses in ECS scenario

kevinrobinson:
e.g. `/teachermoments/ecs?text` per user request, not clear why yet.


kevinrobinson:
selfie

PR #277, opened 2017-06-27
Add key to force replacing component on scene change

kevinrobinson:
This impacts the Smith scenario.

### DIFF
--- a/ui/src/message_popup/renderers/question_interpreter.jsx
+++ b/ui/src/message_popup/renderers/question_interpreter.jsx
@@ -47,6 +47,7 @@ export default React.createClass({
 
     if (question.write) {
       return <MinimalTextResponse
+        key={key}
         forceResponse={true}
         responsePrompt="Notes:"
         recordText="Next"


ui/src/message_popup/renderers/question_interpreter.jsx
```diff
@@ -47,6 +47,7 @@ export default React.createClass({
 
     if (question.write) {
       return <MinimalTextResponse
+        key={key}
```
kevinrobinson: @milsim11 FYI, currently when you advance scenes, if you've scrolled down it will stay scrolled down in the next scene.  This changes incidentally makes each scene show starting at the top, by removing the previous scene altogether and adding a new one each time.  `key` is a special property in React that tells React when a part of the UI should be considered unique, versus when it can be re-used if the contents end up being the same.

This is a bit different than how `key` is normally used, if you want to read more you can check out https://facebook.github.io/react/docs/lists-and-keys.html or https://facebook.github.io/react/docs/reconciliation.html.

kevinrobinson:
selfie

PR #279, opened 2017-06-27
Creates SmithB (which has a different set of scenes for observation round 2)

milsim11:
This branch adds a SmithB simulation. SmithB differs from smithA in that for observation round 2 (within the simulation) smithB adds a different set of 6 scenes instead of reusing the same ones as observation round 1. Note that the addition of SmithB does not, as of yet, make smithA obsolete. 

SmithB_experience_page is the same as SmithA_experience_page with the exception that all 'smithA's are switched to 'smithB's. 

The real changes come within smithB_scenario. As smithB_scenario is based on smithA_scenario the only changes within the first 245 lines are wording changes at line 34, and the addition of more background information on line 71. Post line 245 begins the round 2 scene set and most of it is newly written or heavily modified code. 

newsceneexampleimage

### DIFF
--- a/ui/src/message_popup/playtest/smithB_scenario.jsx
+++ b/ui/src/message_popup/playtest/smithB_scenario.jsx
@@ -1,0 +1,409 @@
+/* @flow weak */
+import React from 'react';
+
+/*
+This file defines the content for the scenario around detecting microaggressions.
+*/
+
+export type QuestionT = {
+  type:string, // Used as a label
+  text:string,
+  open:?bool, // Ask for open-ended user response
+  choices:?bool // Forced-choice response
+};
+
+
+function slidesFor(cohortKey) {
+  const slides:[QuestionT] = [];
+
+  slides.push({ type: 'Overview', el:
+    <div>
+    <div>1. Set Context</div>
+    <div>Imagine yourself as a teacher situated in the context of the particular school, classroom and subject.</div>
+    <br />
+    <div>2. Background</div>
+    <div>A little bit of background information on the lesson being taught.</div>
+    <br />
+    <div>3. Try it!</div>
+    <div>When you're ready, you'll go through a set of short scenes that simulate moments in the classroom. Note what you observe.</div>
+    <br />
+    <div>4. PAUSE!</div>
+    <div>Pause for a brief group discussion before resuming.</div>
+    <br />
+    <div>5. Lenses</div>
+    <div>Try the scenes again, this time from a new perspective.</div>
+  </div>
+  });
+
+
+// Context
+  slides.push({type: 'Context', text: 
+      `Imagine you're a teacher at a suburban high school. Your friend and experienced colleague, Mr. Smith, asks you to sit in and observe his CS class for a day.
+
+The class, taught by Mr. Smith, has 6 students. The demographics of the class are as follows:
+
+      5 Male Students
+      1 Female Student
+      4 White Students (3 Male, 1 Female)
+      2 African American Students (2 Male)`
+  });
+
+
+
+  // -----------------------------------------------------
+  // BACKGROUND
+  // -----------------------------------------------------
+
+  slides.push({type: 'Background', text: 
+      `Mr. Smith will be teaching a class on encoding and sending formatted text. 
+
+In doing so, his students will be working with the standard internet encoding protocol, ASCII. The class will be using pair programming to complete a project developing their own protocol for sending formatted (i.e bold, italicized, etc.) text. 
+
+      `
+    });
+
+  slides.push({type: 'Background', el: 
+<div>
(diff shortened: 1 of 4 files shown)


ui/src/message_popup/playtest/smithB_scenario.jsx
```diff
@@ -0,0 +1,405 @@
...
+    // ---------------------------------
+
+  slides.push({type: 'Set Context - Lenses', el:
+    <div>
+    <div>Mr. Smith comes back to you and asks you to observe another class of his later in the day. It will cover similar subject matter. For this viewing, Mr. Smith has specifically asked you to focus on student engagement and equity. </div>
+    <br />
+    <div>Questions to ask yourself while viewing the scenes may be:</div>
+    <br />
+    <div>- Which students seem to be engaged with the activity?</div>
+    <div>- How are the interactions impacting the learning and confidence of the students involved?</div>
```
kevinrobinson: Words like belonging, inclusiveness, welcoming, might be interesting here too, depending on what particular angle you want to highlight.

ui/src/message_popup/playtest/smithB_scenario.jsx
```diff
@@ -0,0 +1,405 @@
...
+slides.push({type: 'Set Context - Lenses', text:
+`The new class being taught by Mr. Smith also has 6 students. The demographics of the class are as follows:
+
+  4 Male Students
+  2 Female Students
+
+  3 White Students (2 Male, 1 Female)
+  1 African American Student (Female)
+  1 Asian Student (Male)
+  1 Hispanic/Latino Student (Male)
```
kevinrobinson: Just noticing this now (similarly for A), but this seems like a lot of data, particularly with slicing it by different dimensions.

Would a a list of each students' name, with their race, ethnicity and  gender be simpler?

ui/src/message_popup/playtest/smithB_scenario.jsx
```diff
@@ -0,0 +1,405 @@
...
+
+
+
+slides.push({type: 'Try it! - Lenses', el:
+<div>
+<div>Mr. Smith returns the class to discussion:</div>
+<div><img src="https://s3-us-west-2.amazonaws.com/tsl-public/threeflows/smith-scenario/SmithB-Class2.jpg" style={{display: 'block', width: '90%', margin: 20}} /></div>
+<div>Mr. Smith: "So what did you all come up with in your groups? Lee?"</div>
+<br />
+<div>Lee: "Jimmy and I realized that we can use his original idea. We only need numbers zero to nine, which we can map to binary zero to nine. Then, the 26 letters of the alphabet can map to binary 10-35, and the 26 captial letters can map to 36-61."</div>
```
kevinrobinson: This gets more content-heavy with is great for a CS class but might make it harder for playtest folks to parse things like "is that a meaningful contribution to the conversation?"  You could try adding a "background knowledge" section with two sentences at the beginning.

ui/src/message_popup/playtest/smithB_scenario.jsx
```diff
@@ -0,0 +1,405 @@
...
+
+
+
+slides.push({type: 'Try it! - Lenses', el:
+<div>
+<div>Mr. Smith discusses homework with the class:</div>
+<div><img src="https://s3-us-west-2.amazonaws.com/tsl-public/threeflows/smith-scenario/SmithB-class4focus.jpg" style={{display: 'block', width: '90%', margin: 20}} /></div>
+<div>Mr. Smith: "We didn't get as far as I would have liked today, so I will be assigning the rest of the in-class assignment as homework. I just want to make sure that everyone has access to all the necessary resources at home to complete the assignment. Jasmine, Jose. Do you have everything you need?"</div>
+<br />
+<div>Jose: "yeah..."</div>
```
kevinrobinson: capital

ui/src/message_popup/playtest/smithB_scenario.jsx
milsim11: Good idea, added it

ui/src/message_popup/playtest/smithB_scenario.jsx
milsim11: Thanks! I added a bit in the early background. I also added a disclaimer about the tech jargon to the second scene with the suggestion that people not focus on the technical details but the interactions themselves.

ui/src/message_popup/playtest/smithB_scenario.jsx
milsim11: got it

ui/src/message_popup/playtest/smithB_scenario.jsx
kevinrobinson: 👍

PR #281, opened 2017-07-24
Timed Smith

milsim11:
I added a new smith scenario that is the same as smith B with the addition of a timing system for the scene viewing. After 25 seconds the tester is sent to a notes page on which they can note what they see. 

For the app and index pages I simply added the necessary lines for the teachermoments/smith page. 

Only function name changes for the smith experience page (lines 13, 25, 55).

Long text response is a copy of minimal_text_response with changes to field sizes in lines 30 and 77.

Minimal timed response is a highly simplified, and modified form of minimal_text_response. I removed all response aspects from the code. I added a 25 second scene timer in the initialization of the page (line 27) which calls a function to change pages on completion. I also left in the "next" button, and added a remove timer line to 'onSkipTapped' (line 45-46) so that the timer doesn't continue to run and mess up future pages. 

The smith scenario page is copied from smithB scenario. The only notable changes is the switch of scenes over to the timed version (25 second timed view of each scene), and the movement of the note taking to a page after each scene. This is done by a new writeLong slide after each scene slide. 

screen shot 2017-07-24 at 1 41 10 pm

screen shot 2017-07-24 at 1 42 27 pm

For question interpreter I added the importing of both minimal_timed_response and long_text_response. I also added functions for each of these allowing them to be called in the scenario pages.


ui/src/message_popup/renderers/long_text_response.jsx
```diff
@@ -0,0 +1,111 @@
+/* @flow weak */
+import React from 'react';
+
+import TextField from 'material-ui/TextField';
+import RaisedButton from 'material-ui/RaisedButton';
+
+
+// This response supports responding with text.
```
kevinrobinson: Copy and pasting works well for certain components, and for others it's worth figuring out how to change them so that they can just be passed different parameters.  For the experience pages, those are the top-level component so we've been doing more copy-pasting that we might do for a smaller component that is re-used.

It's super helpful how you called out that this component is the same as `MinimalTextResponse` but just longer.  This would be a great case to change `MinimalTextResponse` so that the height could be passed in as a parameter, rather than having to copy and create a new component.  That way all the other details in here don't need to be repeated, and can stay defined in one place.

But it turns out that `MinimalTextResponse` already supports that - you can just pass in `textHeight: 192` or whatever you like.  So you can delete this component, and then update https://github.com/mit-teaching-systems-lab/threeflows/pull/281/files#diff-22c87e71dea0aa864896af1a42a86397R63 to call `MinimalTextResponse` but with a `textHeight` prop.

ui/src/message_popup/renderers/question_interpreter.jsx
```diff
@@ -57,6 +59,28 @@ export default React.createClass({
     }
 
 
+    if (question.writeLong) {
```
kevinrobinson: For this, I'd recommend changing `writeLong` here and in the scenario file to `notes`.  That captures the meaning of what should be happening, and then the details of how that gets shown happen to involve that it has a bigger text box.

ui/src/message_popup/renderers/question_interpreter.jsx
```diff
@@ -57,6 +59,28 @@ export default React.createClass({
...
+        responsePrompt="Notes:"
+        recordText="Next"
+        onLogMessage={onLogMessage}
+        onResponseSubmitted={onResponseSubmitted}
+      />;
+    }
+
+
+    if (question.timed) {
+      return <MinimalTimedView
```
kevinrobinson: This is just the "response" or "interaction" bit, could we rename it to something that indicates that, maybe `TimedAutoAdvanceResponse`?

ui/src/message_popup/renderers/minimal_timed_view.jsx
```diff
@@ -0,0 +1,88 @@
+/* @flow weak */
+import React from 'react';
+import RaisedButton from 'material-ui/RaisedButton';
+
+var sceneTimer;
+
+// This response supports responding with text.
```
kevinrobinson: Stale comment here, please update

ui/src/message_popup/renderers/minimal_timed_view.jsx
```diff
@@ -0,0 +1,88 @@
...
+    return {
+      recordText: 'Respond',
+      autoFocus: true,
+      textHeight: 48
+    };
+  },
+
+
+  getInitialState() {
+    sceneTimer = setTimeout(this.onTimer, 25000);
```
kevinrobinson: Nice!  This is exactly the essence of how to do this. 👍 

There's three small details here about working with timers, which involve a bit more about how React component lifecycles work.

First, instead of using a variable at the top of the file (`var sceneTimer`), let's define this as an instance variable that's part of the React component.  The variable at the top is "closed over" any time this code is run, and so there isn't a unique `sceneTimer` for each `MinimalTimedView` that is created if the page happened to have multiple instances of that component.  That doesn't really matter for our use case, but in general if you have a component or a class, you want to encapsulate its behavior as an instance variable (or in React as state) so that each instance has its own copy.  Concretely, remove the `var` declaration up top, and use `this.sceneTimer = ...` to attach that variable value to the particular React component instance that's created.  You can then read that value back with `this.sceneTimer`.

Second, calling `setTimeout` is a "side effect," meaning that calling it doesn't return a number or a string or a value but it causes something else to happen.  In React, the place for doing that is `componentDidMount` and not `getInitialState`, which should just return a value and not causing anything to happen (like making a server request or printing anything to the console or scheduling any timers).  So just add `componentDidMount` and move this there.

Third, when this component is mounted on the page, the timer starts.  What happens if something else gets rendered instead?  Like, say the user clicks on the reset button in the corner and restarts the scenario altogether?  Right now, nothing would cancel that timer, and it would still run 25 seconds later.  It would then either try to call log message and probably lead to some errors.  You've already taken care of this for the particular case when the user clicks the skip button, but whenever there are "side effects" or timers or other things that we set up in `componentDidMount`, the place to tear those down or clean them up is in `componentWillUnmount`.  That way it'll get cleaned up no matter what happens when this component is unmounted and removed from the screen, however that happens.  So concretely, add `componentWillUnmount` and move the `clearTimeout` into there.  If skip is clicked, it will lead to rendering something different on the page, and then `componentWillUnmount` will be called and the timer will be cleared then.

More on React lifecycle methods here if you're curious: https://facebook.github.io/react/docs/react-component.html.

ui/src/message_popup/renderers/minimal_timed_view.jsx
```diff
@@ -0,0 +1,88 @@
...
+  }
+});
+
+
+const styles = {
+  container: {
+    paddingBottom: 5,
+    margin: 20
+  },
+  textField: {
```
kevinrobinson: I think all of these styles except `container` can be deleted?  Just check if they're used in this file and if they're not used anymore delete 'em!

ui/src/message_popup/renderers/minimal_timed_view.jsx
```diff
@@ -0,0 +1,88 @@
...
+
+
+  getInitialState() {
+    sceneTimer = setTimeout(this.onTimer, 25000);
+    return {
+      responseText: 'type here!'
+    };
+  },
+
+  onTextChanged(e) {
```
kevinrobinson: This isn't used anymore, can delete it!

ui/src/message_popup/renderers/minimal_timed_view.jsx
```diff
@@ -0,0 +1,88 @@
...
+      autoFocus: true,
+      textHeight: 48
+    };
+  },
+
+
+  getInitialState() {
+    sceneTimer = setTimeout(this.onTimer, 25000);
+    return {
+      responseText: 'type here!'
```
kevinrobinson: This piece of state isn't used in this component, this is left over from copy-pasting.  You can delete this (and with the other comment above) you can remove `getInitialState` altogether.

ui/src/message_popup/renderers/minimal_timed_view.jsx
```diff
@@ -0,0 +1,88 @@
...
+  },
+
+  onTextChanged(e) {
+    const responseText = e.target.value;
+    this.setState({responseText});
+  },
+
+  onTimer() {
+    const {responseText} = this.state;
+    this.props.onLogMessage('message_popup_text_response', {responseText});
```
kevinrobinson: This is recording something slightly different than a text response.  Let's change the `message_popup_text_response` message type to match whatever the component name is (eg., `timer_auto_advance_response`.  Also, `responseText` is meaningless here (the user isn't typing anything in this component) so we can remove those bits too.

ui/src/message_popup/renderers/minimal_timed_view.jsx
```diff
@@ -0,0 +1,88 @@
...
+    this.props.onResponseSubmitted({ignore: true});
+  },
+
+
+  onSkipTapped() {
+    clearTimeout(sceneTimer);
+    const {responseText} = this.state;
+    this.props.onLogMessage('message_popup_text_response', {responseText});
```
kevinrobinson: We should update the message type to `timer_auto_advance_response` or something similar (same as above).  Also, we probably want to capture in the data when people are skipping versus getting forced along.  So here you could add some data in the second argument:

```
this.props.onLogMessage('timer_auto_advance_response', {skipTapped: true});
```

That way we can tell looking in the data on the server if we want to know how often people are skipped versus getting forced along.

ui/src/message_popup/renderers/minimal_timed_view.jsx
milsim11: Got it!

ui/src/message_popup/renderers/minimal_timed_view.jsx
milsim11: done

ui/src/message_popup/renderers/minimal_timed_view.jsx
milsim11: I think this has been fixed

ui/src/message_popup/renderers/minimal_timed_view.jsx
milsim11: done

ui/src/message_popup/renderers/minimal_timed_view.jsx
milsim11: sounds good

ui/src/message_popup/renderers/minimal_timed_view.jsx
milsim11: pretty sure I got this fixed up

ui/src/message_popup/renderers/minimal_timed_view.jsx
milsim11: done

ui/src/message_popup/renderers/question_interpreter.jsx
milsim11: done

ui/src/message_popup/renderers/question_interpreter.jsx
milsim11: switched

ui/src/message_popup/renderers/long_text_response.jsx
milsim11: fixed. Thanks, didn't realize that

PR #282, opened 2017-07-25
Adds rosa counseling Scenario

milsim11:
Adds the rosa counseling scenarios to teacher moments under teachermoments/rosa.

The code is based off of the jayden scenario. The only changes for the experience page is the replacement of all "Jaydens" (lines 13, 25, 55) with "rosa's". The rosa_scenario page follows the form of the jayden scenario but with new text. 
screen shot 2017-07-25 at 2 57 34 pm

### DIFF
--- a/ui/src/message_popup/playtest/rosa_experience_page.jsx
+++ b/ui/src/message_popup/playtest/rosa_experience_page.jsx
@@ -1,0 +1,132 @@
+/* @flow weak */
+import React from 'react';
+import uuid from 'uuid';
+
+import * as Api from '../../helpers/api.js';
+import hash from '../../helpers/hash.js';
+import LinearSession from '../linear_session/linear_session.jsx';
+import SessionFrame from '../linear_session/session_frame.jsx';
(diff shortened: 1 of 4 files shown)


ui/src/message_popup/playtest/rosa_experience_page.jsx
```diff
@@ -0,0 +1,132 @@
...
+import AudioResponseSummary from '../renderers/audio_response_summary.jsx';
+
+type ResponseT = {
+  choice:string,
+  question:QuestionT
+};
+
+
+
+// This is a scenario around bubble sort and classroom management.
```
kevinrobinson: Since this is exactly the same as Jayden (just different questions), and is likely to stay the same overall structure, this could be a good abstraction to make - something like a `CounselingScenario` component that just takes the set of questions to use.  Then the `JaydenExperiencePage` and `RosaExperiencePage` both call that but provide different sets of questions.

If you're into that, you could do that in this pull request, or it's fine to leave this as its own file that's copy-pasted for now too.

PR #286, opened 2017-07-27
Update QuestionInterpreter to allow forcing text, enable for Rosa

kevinrobinson:
This updates `QuestionInterpreter` to allow a prop forcing text responses, aimed at supporting folks who have audio recording capabilities but no microphone (the current check for audio recording capabilities can't detect this).

It adds a `text` query string to the Rosa scenario passing this through.


kevinrobinson:
selfie

PR #287, opened 2017-07-27
Polyfill Promise for IE

kevinrobinson:
In an older version of IE, Promise is undefined and so this leads JS execution to stop on boot - no page is rendered and the error isn't reported to Rollbar.  This polyfills that.


kevinrobinson:
selfie

kevinrobinson:
via Browserstack the only way I could reproduce was on a Window phone, but manually tested this with an ECS pal on his computer now and it works as expected.

PR #291, opened 2017-08-01
Adds a timed version of Smith

milsim11:
I added a new smith scenario that is the same as smith B with the addition of a timing system for the scene viewing. After 25 seconds the tester is sent to a notes page on which they can note what they see.

For the app and index pages I simply added the necessary lines for the teachermoments/smith page.

Only function name changes for the smith experience page (lines 13, 25, 55).

Minimal timed view is a highly simplified, and modified form of minimal_text_response. I removed all response aspects from the code. I added a 25 second scene timer in a componentDidMount function, and the clearTimeout function in an componentWillUnmount function. I also left in the "next" button,  so that the user can skip forward when completed.

The smith scenario page is copied from smithB scenario. The only notable changes is the switch of scenes over to the timed version (25 second timed view of each scene), and the movement of the note taking to a page after each scene. 

screen shot 2017-07-24 at 1 41 10 pm

screen shot 2017-07-24 at 1 42 27 pm

For question interpreter I added the importing of minimal_timed_response. I also added a function for  this allowing it to be called in the scenario pages.


ui/src/message_popup/playtest/smith_experience_page.jsx
```diff
@@ -0,0 +1,129 @@
...
+  },
+
+
+  renderIntro() {
+    return (
+      <IntroWithEmail defaultEmail={this.state.email} onDone={this.onStart}>
+        <div>
+          <p>Welcome!</p>
+          <p>This is an interactive case study simulating a small part of a high school computer science lesson.</p>
+          <p>You'll review the context of the lesson briefly, share what you anticipate about the lesson, and then try it out!  Afterward you'll reflect before heading back to debrief with the group or share online.</p>
```
kevinrobinson: I missed this before, but does this copy need to be updated?

ui/src/message_popup/renderers/minimal_timed_view.jsx
```diff
@@ -0,0 +1,68 @@
...
+  componentWillUnmount(){
+    clearTimeout(this.sceneTimer);
+  },
+
+  sceneTimer: (null: ?number),
+  
+
+  onTimer() {
+    this.props.onLogMessage('timer_auto_advance_response');
+    this.props.onResponseSubmitted({ignore: true});
```
kevinrobinson: You can remove `{ignore: true}`, that was leftover from code that would help tell the difference between when the user tapped the "Respond" or "Move on" buttons.

ui/src/message_popup/renderers/minimal_timed_view.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+
+  render() {
+    const {
+      recordText
+    } = this.props;
+   
+    return (
+      <div className="MinimalTimedView" style={styles.container}>
+        <div style={styles.buttonRow}>
```
kevinrobinson: This is referencing `styles.buttonRow`, but looks like that's been deleted?  If this looks the way you want visually on the screen, then you can just remove the `style={...}` bit.

ui/src/message_popup/renderers/question_interpreter.jsx
```diff
@@ -57,6 +58,29 @@ export default React.createClass({
...
+        forceResponse={true}
+        responsePrompt="Notes:"
+        recordText="Next"
+        onLogMessage={onLogMessage}
+        onResponseSubmitted={onResponseSubmitted}
+      />;
+    }
+
+
+    if (question.TimedAutoAdvance) {
```
kevinrobinson: This is great, but the other values (eg., `open` or `notes` or `choices`) are all camelCase (starting with lowercase, no space between and uppercase to start new words).  So please change this to `timedAutoAdvance` to match.

milsim11: done

ui/src/message_popup/renderers/minimal_timed_view.jsx
milsim11: this caused the timer to stop running, so I'm not gunna mess with it.

ui/src/message_popup/playtest/smith_experience_page.jsx
milsim11: Good catch, fixed it.

PR #293, opened 2017-08-04
Add HMTCA with reviewing page, endpoints for reviewing, add server linting

kevinrobinson:
This adds a new `/teachermoments/hmtca` route as a first prototype for the HMTCA session.  It also adds two new server endpoints - each for writing and reading `anonymized_apples_to_apples` data storing `(apples_key, scene_number, scene_text, anonymized_text)` so that we can store this data separately from data with other identifying data (eg., sessionId, timestamp).  These are used in a review UI at the end.  There's no way to synchronize between different users at the moment - that would be a great use for adding in a timer loop :)

In the process, I ran into some linting errors and also added linting for the server code.

screen shot 2017-08-04 at 1 53 02 pm
screen shot 2017-08-04 at 2 43 11 pm

screen shot 2017-08-04 at 1 10 43 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "watch": "cd ui && npm run watch && cd ..",
     "heroku-prebuild": "cd ui && mkdir build && npm install",
     "heroku-postbuild": "cd ui && npm run build",
-    "test": "cd ui && npm run test"
+    "lint": "eslint --ext jsx --ext js -c server/.eslintrc server",
+    "test": "npm run lint && cd ui && npm run test"
   },
   "dependencies": {
     "aws-sdk": "^2.4.8",
@@ -30,6 +31,8 @@
   "license": "MIT",
   "devDependencies": {
     "babel-plugin-react-transform": "^2.0.2",
+    "eslint": "^4.3.0",
+    "eslint-plugin-node": "^5.1.1",
     "livereactload": "^2.2.3",
     "react-proxy": "^1.1.8"
   }
--- a/server/endpoints/apples.js
+++ b/server/endpoints/apples.js
@@ -1,0 +1,55 @@
+// Get all sorted responses that fit this particular type for a particular
+// `applesKey`.  This only pulls out the anonymized text responses.
+function getAnonymizedApples(params, cb) {
+  const {queryDatabase, applesKey} = params;
+
+  const sql = `
+    SELECT
+      json->'anonymizedText' as anonymized_text,
+      json->'sceneNumber' as scene_number,
+      json->'sceneText' as scene_text
+    FROM evidence
+    WHERE 1=1
+      AND type = 'anonymized_apples_to_apples'
+      AND json->>'applesKey' = $1
+    ORDER BY
+      scene_number ASC,
+      scene_text ASC,
+      anonymized_text ASC;`;
+  const values = [applesKey];
+  queryDatabase(sql, values, (err, result) => {
+    if (err) {
+      console.log('getAnonymizedApples: error', JSON.stringify(err));
+      return cb(err);
+    }
+
+    const {rows} = result;
+    return cb(null, {rows});
+  });
+}
+
+
+module.exports = {
+  // Queries database for anonymized responses to a particular Apples-to-Apples key
+  sensitiveGetApples({queryDatabase}) {
+    return (request, response) => {
+      const applesKey = request.params.key;
+      console.log('sensitiveGetApples: applesKey = ' + applesKey);
+      getAnonymizedApples({queryDatabase, applesKey}, (err, result) => {
+        if (err) {
+          console.log('sensitiveGetApples: error', JSON.stringify(err));
+          response.status(500);
+          response.json({ status: 'error' });
+          return;
+        }
+
+        const {rows} = result;
+        console.log(`sensitiveGetApples: returning ${rows.length} rows...`);
+        response.status(200);
+        response.json(rows);
+        return;
+      });
+    };
+  }
+};
+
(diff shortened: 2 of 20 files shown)


package.json
```diff
@@ -11,7 +11,8 @@
     "watch": "cd ui && npm run watch && cd ..",
     "heroku-prebuild": "cd ui && mkdir build && npm install",
     "heroku-postbuild": "cd ui && npm run build",
-    "test": "cd ui && npm run test"
+    "lint": "eslint --ext jsx --ext js -c server/.eslintrc server",
```
kevinrobinson: This is adding commands to run linting on the server code too, I thought that was already set up but wasn't.

server/endpoints/apples.js
```diff
@@ -0,0 +1,55 @@
...
+
+    const {rows} = result;
+    return cb(null, {rows});
+  });
+}
+
+
+module.exports = {
+  // Queries database for anonymized responses to a particular Apples-to-Apples key
+  sensitiveGetApples({queryDatabase}) {
```
kevinrobinson: This is the server code for reading back the data for anonymized responses.

ui/src/message_popup/playtest/hmtca_experience_page.jsx
```diff
@@ -0,0 +1,203 @@
...
+    });
+  },
+
+  onResetSession() {
+    this.setState(this.getInitialState());
+  },
+
+  onLogMessage(type, response) {
+    const {cohortKey, bucketId, sessionId, questionsHash} = this.state;
+
```
kevinrobinson: This listens for particular log messages, and then writes them to two different places on the server - one the usual place with all the sessionId and timestamp data, and another with just the info about scenes and anonymized text responses.

ui/src/message_popup/playtest/hmtca_experience_page.jsx
```diff
@@ -0,0 +1,203 @@
...
+    return <LinearSession
+      questions={questions}
+      questionEl={this.renderQuestionEl}
+      summaryEl={this.renderClosingEl}
+      onLogMessage={this.onLogMessage}
+    />;
+  },
+
+
+  renderIntro() {
```
kevinrobinson: This is a different intro screen, based on the prototype deck.  The value the user picks here gets passed along into the code returning the slides in the `hmtca_scenario.jsx` file.

ui/src/message_popup/playtest/hmtca_group_review.jsx
```diff
@@ -0,0 +1,103 @@
...
+    };
+  },
+
+  componentDidMount() {
+    this.refreshResponses();
+  },
+
+  refreshResponses() {
+    const {applesKey} = this.props;
+    Api.getApples(applesKey).end(this.onResponsesReceived);
```
kevinrobinson: This loads the responses from the server based on: 1) the code they put in at the beginning, and 2) the bucket they chose.  So if they used the same code and same bucket and kept playing day-after-day, they'd keep seeing more and more responses in the review screen.

ui/src/message_popup/playtest/hmtca_group_review.jsx
```diff
@@ -0,0 +1,103 @@
...
+    const applesResponses = JSON.parse(response.text);
+    this.setState({ hasLoaded: true, applesResponses });
+  },
+
+  render() {
+    const {hasLoaded, applesResponses} = this.state;
+    if (!hasLoaded) return <div>Loading...</div>;
+
+    // Use scene number for ordering, but then just number then naturally
+    const sceneToResponses = _.groupBy(applesResponses, 'scene_number');
```
kevinrobinson: This sorts out the responses by scene, and then goes through each scene in order, and then makes a sticky-note-looking card for each anonymized response.

ui/src/message_popup/playtest/hmtca_scenario.jsx
```diff
@@ -0,0 +1,43 @@
...
+  { id: 202, text: 'Disrespect towards teachers' },
+  { id: 203, text: 'Disrespect towards females' },
+  { id: 204, text: 'Combination of the above' }
+];
+
+
+function slidesFor(cohortKey, bucketId) {
+  const slides:[QuestionT] = [];
+
+  if (bucketId === 201) {
```
kevinrobinson: These are mostly placeholder scenes from now, but this shows how we can have different scenes based on the buckets they chose, and how to number them so the responses get tracked for reviewing later (eg., if you put a `note` or `open` or other type of question in these slides that's cool but they won't show up in the review at the end).

ui/src/message_popup/renderers/apples_text_response.jsx
```diff
@@ -0,0 +1,48 @@
...
+    responsePrompt: React.PropTypes.string,
+    recordText: React.PropTypes.string,
+    ignoreText: React.PropTypes.string,
+    autoFocus: React.PropTypes.bool,
+    textHeight: React.PropTypes.number
+  },
+
+
+  // Watch for text responses, and double-log them.
+  onApplesLogMessage(type, response) {
```
kevinrobinson: This wraps the `MinimalTextResponse` component, listens for some data it broadcasts out, and then adds in some other data that's needed about the scene number or whether the user clicked "Move on" so we can show that in the review at the end as well.

kevinrobinson:
@milsim11 Here's a first cut at some pieces of this.  I left some comments of places that might be helpful as you're reading through, and we can talk through this tomorrow, or you could pull down this branch and try it out, or we could get it merged tomorrow morning, whatever you think works best.

kevinrobinson:
selfie cc @milsim11

PR #294, opened 2017-08-07
adds facilitated and fully online options to the smithTimer scenario

milsim11:
This branch takes smithTimer and splits it into two web-addresses (each with subtle scenario differences) based on whether the scenario is being facilitated or taken fully online. 

The two new web-addresses are as follows:
     /teachermoments/smithFacilitated:  For facilitated uses of the Smith scenario
     /teachermoments/smith :  For fully online (individual) use of the Smith scenario

Code Modifications:

App.jsx    -  Adds the two URLs for the smith scenario. Depending on which is accessed, a boolean (for use within the singular smith_scenario.jsx) is set to represent whether it is a facilitated or individual use.

smith_experience_page.jsx    -   Adds a required facilitated/unfacilitated boolean. Sets default boolean value to false. Changes, allQuestions to pass said boolean to smith_scenario. 

smith_scenario.jsx    -  Adds boolean variable to slidesFor function. Lines 32-39 change Step 4 depending on whether the scenario is being facilitated or not.  Made cosmetic changes to the notes pages. 290-345 changes/adds the post scene set #1 questions depending on facilitated (this is a major change). Similarly, lines 538-575 do the same for scene set #2 (the other major change). 

screen shot 2017-08-07 at 9 57 43 am

### DIFF
--- a/ui/src/message_popup/playtest/smith_experience_page.jsx
+++ b/ui/src/message_popup/playtest/smith_experience_page.jsx
@@ -28,12 +28,23 @@ export default React.createClass({
     query: React.PropTypes.shape({
       cohort: React.PropTypes.string,
       p: React.PropTypes.string
-    }).isRequired
+    }).isRequired,
+    facilitated: React.PropTypes.bool.isRequired
   },
 
   contextTypes: {
     auth: React.PropTypes.object.isRequired
   },
+
+
+
+
+  getDefaultProps() {
+    return {
+      facilitated: false
+    };
+  },
+
 
   // Cohort comes from URL
   getInitialState() {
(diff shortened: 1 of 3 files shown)


ui/src/message_popup/playtest/smith_experience_page.jsx
```diff
@@ -52,8 +63,8 @@ export default React.createClass({
   // Making questions from the cohort
   onStart(email) {
     const {cohortKey} = this.state;
-    const allQuestions = smithScenario.questionsFor(cohortKey);
-
+    const {facilitated} = this.props;
+    const allQuestions = smithScenario.questionsFor(cohortKey, (facilitated));
```
kevinrobinson: Passing facilitated here is good, and you're reading it in on the other end correctly too.  In general, passing boolean values can be hard to understand, and for options like this it's usually more explicit to pass it as an object that says what the boolean value is.

So here you could change this to `{isFacilitated: facilitated}` and then update the two places in the scenario code that check for this so that they take `options` and then they check for `options.isFacilitated`.

ui/src/message_popup/playtest/smith_experience_page.jsx
```diff
@@ -28,13 +28,24 @@ export default React.createClass({
...
   },
 
   contextTypes: {
     auth: React.PropTypes.object.isRequired
   },
 
+
+
+
+  getDefaultProps() {
```
kevinrobinson: You don't need this since you set this prop to `isRequired` up above in the `propTypes`.  This is only for if the prop is optional.

PR #296, opened 2017-08-08
Add anonymous identifier to start HMTCA session

kevinrobinson:
Allow folks to add an anonymous identifier to join with paper consent and survey data.

screen shot 2017-08-08 at 7 11 29 am


kevinrobinson:
selfie cc @milsim11

PR #299, opened 2017-08-08
adds refresh timer, and pause slide

milsim11:
@kevinrobinson Here is the timer refresh branch. It also has the pause screen.


ui/src/message_popup/playtest/hmtca_group_review.jsx
```diff
@@ -1,4 +1,4 @@
-/* @flow weak */
+// TODO(kr) flow typing disabled because of error in refreshTimer typing
```
kevinrobinson: @milsim11 This is a hack workaround for now since we have 10 minutes :)  This removes the annotation at the top that tells Flow to type check this file.

kevinrobinson:
@milsim11 I worked around the type error and resolved the merge conflict here.  Merging now since you're setting up other stuff!

PR #301, opened 2017-08-14
PLC restructuring

milsim11:
This update removes the option to choose your scene topic, and leaves only 4 of the original 18 scenes. It also includes rewrites of all the instruction pages, and changes the slide arrangements so that the instructions for each part comes directly before their respective parts. 

hmtca_experience_page:
- Starts by removing now unused variables (connected to topic selection which was removed).
- Removal of lines 197-214 is the removal of the scene selection process.
- Minor wording changes...
- Removes all requirements for BucketID variable

hmtca_scenario:
- Removes extra bucket IDs
- Adds more team codes (now at 26)
- Changes to instruction set
- Removes excess scenes and leaves only the 4 scenes users will run in new iteration(Lines 62-155) 

image

image


milsim11:
WOW Kenny this is so good! (And by "so good" I mean "good enough for now, we can streamline it later"). 
image

milsim11:
Thanks other Kenny! Shippin it! SELFIE! 
image

kevinrobinson:
👍

PR #304, opened 2017-08-21
Add workshop as part of applesKey for per-session isolation

kevinrobinson:
This allows passing in a `workshop` value that is part of the `applesKey` used for storing values that are reviewed.

For workshops, we can set the URL we give folks to include `workshop=foo` as part of the query string, to ensure that all folks within a single workshop all use the same value and all can review each others' responses.


kevinrobinson:
@milsim11 This is slightly different than clearing old responses, but accomplishes the same thing while allowing us to keep all the old responses too.

kevinrobinson:
@milsim11 and I reviewed together in person

PR #305, opened 2017-08-21
Add in lettering to responses when reviewing

kevinrobinson:
So that folks can talk about responses using letter names instead of "the third one from the bottom".

Screenshot:
screen shot 2017-08-21 at 11 29 47 am


kevinrobinson:
FYI @milsim11 this is what we looked at together in person

kevinrobinson:
paired with @milsim11

PR #307, opened 2017-08-21
Add minimal tests to insubordination scenarios

kevinrobinson:
Used in unsuccessfully debugging a Rollbar error, adding this in for a bit more coverage.


kevinrobinson:
selfie

PR #308, opened 2017-08-22
Add feature for reviewing notes when giving feedback in Smith scenario

kevinrobinson:
This allows folks to review their past notes when giving feedback in the Mr. Smith scenario.

It works by modifying the scenario to have `feedback` questions, and `getNextQuestion` then checks for this and grabs the past notes to include them in the question when rendering.  Rendering is done by `ResponseWithPastNotes`.

Screenshot:
screen shot 2017-08-22 at 10 13 45 am

### DIFF
--- a/ui/src/message_popup/playtest/smith_experience_page.jsx
+++ b/ui/src/message_popup/playtest/smith_experience_page.jsx
@@ -1,6 +1,7 @@
 /* @flow weak */
 import React from 'react';
 import uuid from 'uuid';
+import _ from 'lodash';
 
 import * as Api from '../../helpers/api.js';
 import hash from '../../helpers/hash.js';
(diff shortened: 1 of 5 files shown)


ui/src/message_popup/playtest/smith_experience_page.jsx
```diff
@@ -60,6 +61,32 @@ export default React.createClass({
...
+    }
+
+    return question;
+  },
+
+  // Returns an array of past notes for review as part of a feedback question.
+  // This looks at all `notes` responses prior to `feedbackQuestionIndex`, only 
+  // looking back until a previous `feedback` question, since there may be multiple
+  // in the same scenario.
+  getNotesBeforeFeedbackIndex(questions, responses, feedbackQuestionIndex) {
```
kevinrobinson: @milsim11 I got mixed up writing this function initially, so wrote a test to verify it's working correctly.  You can see that in `smith_experience_page_test.jsx`, which adds a fake scenario and verifies that the method does what we'd expect.  This is similar to testing manually, but is checked into source control and can run automatically when we make changes and faster too!

kevinrobinson:
@milsim11 How does this look to you?

kevinrobinson:
selfie, other changes came in that @milsim11 is working on now

PR #310, opened 2017-08-23
Allow forcing text by URL in Jayden scenario

kevinrobinson:
So you can add `?text` to force this if that's what we need for a particular session.


kevinrobinson:
selfie

PR #312, opened 2017-08-24
Add param to not show consent page for Smith scenario, if it's alread…

kevinrobinson:
…y been done in the session


kevinrobinson:
@milsim11 and I did this in person

PR #313, opened 2017-08-24
Remove reset button from SessionFrame navigation bar

kevinrobinson:
This doesn't serve any purpose, other than people clicking on it thinking that it will replay the scene and getting confused.

screen shot 2017-08-24 at 3 14 43 pm


kevinrobinson:
selfie

PR #314, opened 2017-08-24
Patch Jayden summary to work with text, rename component, fix copy bug

kevinrobinson:
The `AudioResponseSummary` component actually works for text too, so rename it.  There's two kinds of responses that are text, so this updates it to include both (depending on whether it's `MinimalTextComponent` directly or if that component is wrapped by `MinimalOpenResponse`.  Adds a test too.

Screenshot:
screen shot 2017-08-24 at 3 00 25 pm


kevinrobinson:
selfie

PR #315, opened 2017-08-24
Add tests to start fixing two types of text responses

kevinrobinson:
Tracking down why the difference is, and adding tests.  These should be unified to be the same format; that'd be next.


kevinrobinson:
selfie

PR #316, opened 2017-08-30
Revise Jayden slide to 'look like me' to highlight race more directly

kevinrobinson:
The context leading up suggests he might mean singers, this is an attempt to have him more directly bring up that it's about his racial identity.


kevinrobinson:
selfie

PR #317, opened 2017-09-11
Add cross-env to support Windows development

kevinrobinson:
To support @keving17 doing awesome work on his Windows computer!


kevinrobinson:
selfie cc @keving17

PR #321, opened 2017-11-14
Move to yarn and single `start` command

kevinrobinson:
This updates both the server and UI code to use use `yarn` instead of `npm` and checks in the `yarn.lock` files.

It also changes the way to start the project so that there's a single command to start both the server and client in parallel, aiming to be similar to https://github.com/mit-teaching-systems-lab/swipe-right-for-cs, and to incrementally move so the product development experience is the same as for `create-react-app` (that isn't migrated here).


kevinrobinson:
@keving17 This changes things to use `yarn` and to change the way the project is started locally.  Can you review and see if this looks good?

kevinrobinson:
@keving17 FYI merging this now

PR #322, opened 2017-12-01
Home page for equity workshop

kevinrobinson:
This is for folks to land on and do some jigsawing during a workshop.

Example:
screen shot 2017-11-30 at 6 11 18 pm

This makes some minimal changes to other pages to change some workshop-specific language, copies a separate `climate` scenario adapted from the HMTCA work, and updates the `/demos` pages as well:

screen shot 2017-11-30 at 6 16 31 pm


kevinrobinson:
selfie

PR #330, opened 2017-12-20
Fix but with passing onLog in particular prototype pages

kevinrobinson:
Resolving https://rollbar.com/threeflows/threeflows/items/139/?item_page=0&#instances for prototypes.


kevinrobinson:
selfie

PR #334, opened 2018-01-03
Move LikertResponse into responses, add test

kevinrobinson:
Refactoring to move all responses into same folder and capture response types.


kevinrobinson:
selfie

PR #335, opened 2018-01-03
Move TimedAutoAdvanceResponse, ResponseWithPastNotes and factor out withContext

kevinrobinson:
This moves `TimedAutoAdvanceResponse` and `ResponseWithPastNotes ` into the `responses/` folder and adds minimal tests, in one case using `ReactTestRenderer`.  It also factors out the `withContext` helper function.


kevinrobinson:
selfie

PR #336, opened 2018-01-09
Equity page: Update text for themes to broader audience

kevinrobinson:
So they're more applicable to a general audience.


kevinrobinson:
selfie

PR #337, opened 2018-01-12
Add responses to questionEl in LinearSession to fix bug with Turner scenario

kevinrobinson:
This fixes a bug with the Turner scenario that was likely introduced in https://github.com/mit-teaching-systems-lab/threeflows/pull/270.  This bug also affects the Danson scenario, which isn't fixed here and needs to be addressed separately.

The problem comes from two pieces of code.  The first piece of code is in `LinearSession` in https://github.com/mit-teaching-systems-lab/threeflows/pull/270 that merges in the `question` to every response that is logged.  This is useful for some data analysis, since the question that the user was shown is guaranteed to be in the same log data as the response itself (otherwise analysis would have to look back over multiple log items and join that, which is less simple).  The second piece of code is the implementation of `getNextQuestion` in the Turner (and Danson) scenario.  The Turner scenario has some postreflection questions that show the user all of the past questions and their responses.  To accomplish this, we'd previously implement `getNextQuestion` and merged in all past responses to each question.  This was a workaround of not having access to past `responses` in `questionEl`.

These two things interacted so that for the Turner scenario, `getNextQuestion` would put all past responses into the `question`, and then when any logging happened, `LinearSession` would put the `question` into each `response`.  This led to essentially doubling the size of the response that was logged each time, and led to payload too large errors on the server when parsing JSON, and memory errors on the client that could crash the browser.

### DIFF
--- a/ui/src/message_popup/linear_session/linear_session.jsx
+++ b/ui/src/message_popup/linear_session/linear_session.jsx
@@ -78,7 +78,7 @@ export default React.createClass({
 
     return (
       <div key={question.id}>
-        {this.props.questionEl(question, this.onLogMessageWithQuestion.bind(this, question), this.onResponseSubmitted.bind(this, question))}
+        {this.props.questionEl(question, this.onLogMessageWithQuestion.bind(this, question), this.onResponseSubmitted.bind(this, question), {responses})}
       </div>
     );
   }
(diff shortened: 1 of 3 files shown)


kevinrobinson:
@keving17 and I paired on this and did it together 🚢

PR #338, opened 2018-02-06
Convert from Mocha to Jest

kevinrobinson:
This is ahead of @keving17's awesome work to switch the build process over to use `create-react-app`.

Before:
screen shot 2018-02-06 at 4 33 15 pm

After:
screen shot 2018-02-06 at 4 33 31 pm


kevinrobinson:
Looked at this in person with @keving17.  mergin'

PR #339, opened 2018-02-07
New build process for Teacher Moments using create-react-app

keving17:
This changes the build process for Teacher Moments to use create-react-app which should be more compatible with future build processes for react applications moving forward.

Some things to check still
- [x] does this work with CSV and CSS files?
- [x] what is the deal with #! in url
- [x] merge in testing changes from https://github.com/mit-teaching-systems-lab/threeflows/pull/338

PR #340, opened 2018-02-13
Fixup linter errors and Jest setup for create-react-app build

kevinrobinson:
This is intended to be merged into https://github.com/mit-teaching-systems-lab/threeflows/pull/339.

# dependencies
There were missing dependencies that I `yarn add`ed.

# linter
status: all the errors that are left being real actual errors:
screen shot 2018-02-13 at 6 17 15 pm

This involved fixing `react/jsx-no-target-blank` warnings (find-replace following the suggestions about this security bug), and doing some of the other things discussed in https://github.com/mit-teaching-systems-lab/threeflows/pull/339#issuecomment-365438711.

# Jest tests
status: done except for two tests.
screen shot 2018-02-13 at 6 21 33 pm

This didn't work right because https://github.com/mit-teaching-systems-lab/threeflows/pull/338 added jest as a dependency and called it directly.  This updates it to call `react-scripts test` so that it uses all the `create-react-app` config now.  This also involved a few other changes like renaming all the `_test.jsx` files to be `.test.js` so that the create-react-app Jest setup picks them up.  There were some files in the `test` folder that didn't get migrated over from the `ui` folder, so this PR copies them over too.

To finish off https://github.com/mit-teaching-systems-lab/threeflows/pull/339, these two two test will need to be fixed or removed (they're not for high priority pages).

# flow
status: `618 errors`
This is a lot of failure, but I think it's only within ~10 files or so.  The codemod seems to either added type definitions in a way that leads to more type failures for state changes in top-level pages.  This would need more work to finish off https://github.com/mit-teaching-systems-lab/threeflows/pull/339.

This PR resolves:
- `$FlowFixMe` values that the codemods put in.  I updated these manually where possible.
- adding in `.flowconfig`
- Adding type declarations for Jest, which weren't migrated over originally in https://github.com/mit-teaching-systems-lab/threeflows/pull/339


kevinrobinson:
@keving17 check this out and if this looks good and helps out, you can merge it into https://github.com/mit-teaching-systems-lab/threeflows/pull/339.

I won't be able to look at this more this week, but can probably catch up asycn next week.

PR #342, opened 2018-03-01
Fix Flow errors for paper_prototype_page.jsx

kevinrobinson:
@keving17 I looked at the Flow errors a bit more, sorry I didn't get to this yesterday.  The root problem is that the way Flow typing works for React components changed in 0.53 ([docs](https://github.com/facebook/flow/blob/master/Changelog.md#0530)), and we pulled in that upgrade as well in switching to `create-react-app`.  I ran the [flow-upgrade](https://github.com/facebook/flow/tree/5aaf9b9177514c7ed0c2b65af7efff4fea5c5775/packages/flow-upgrade) tool they recommended but it didn't fix anything up automatically.

Fortunately, even though there's 500 some errors, from scanning I think almost all of them are from one particular change, and that this is both a simple and beneficial change to make that only affects ~10-20 files.  Also, accidentally I upgraded from 0.65 to 0.66 in the process of looking at that and they have a more helpful format for error messages too :)

Here's what the last error I see on new-build running flow 0.66:
screen shot 2018-03-01 at 7 21 05 am

And here's one more up a bit further for the same file, `paper_prototype_page.jsx`, so we stay focused on the changes for one file at a time:
screen shot 2018-03-01 at 7 22 59 am

They're both complaining because we're calling `setState` in an unsafe way, where we're trying to set a value for state that may not be type safe.  This is happening because we haven't added type information for the component and Flow now expects this as of 0.53.  So before our component looked like this:
screen shot 2018-03-01 at 7 22 38 am

So Flow can't infer all the types from just that and the new recommended way to do this is to add explicit type for props and state, which is great.  Here's what that looks like:

screen shot 2018-03-01 at 7 16 15 am

In the process of doing this, the Flow error message also correctly pointed out that we were using `questionsHash` in some code but it wasn't in the type definition.  So I added that in.

So to back up, I think this is the root cause of all of these, and that there's probably only 10-15 files to do this for.  In the process we'll also get better type checking on these kinds of mistakes.  So I'd say let's try to fix forward from this and get things working, and I'm happy to spend time on this next Wednesday as well and try to burn through them.  If you want to take this and run with it and try to work through the other files that's awesome but I'm happy to take a crack at that Wednesday too if it's better for you to focus on other week.  If we're still stuck on this next week then maybe we shift strategies.

How does that sound?

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "clipboard": "^1.6.1",
+    "flow-bin": "^0.66.0",
     "material-ui": "^0.20.0",
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
@@ -33,7 +34,6 @@
     "chai": "^4.1.2",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
-    "flow-bin": "^0.65.0",
     "sinon": "^4.3.0"
   }
 }
--- a/client/src/auth_container.test.js
+++ b/client/src/auth_container.test.js
@@ -1,5 +1,5 @@
 /* @flow weak */
-import React from 'react';
+import * as React from 'react';
 
 import {shallow} from 'enzyme';
 import {expect} from 'chai';
--- a/client/src/message_popup/equity/paper_prototype_page.jsx
+++ b/client/src/message_popup/equity/paper_prototype_page.jsx
@@ -36,19 +36,29 @@ const Parts = {
   FINAL_INSTRUCTIONS: 'FINAL_INSTRUCTIONS'
 };
 
+type Props = {
+  prototypeKey: string,
+  query: {
+    cohort?: string,
+    p?: string,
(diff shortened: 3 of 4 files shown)


keving17:
Hey Kevin! Sorry i didn't really get a chance to look too deeply into this. I had begun checking things out before I got caught up in class work. It seems like from https://github.com/mit-teaching-systems-lab/threeflows/pull/343 this is not quite the quick fix you had imagined? Should I just leave this alone for now?

kevinrobinson:
@keving17 Yeah, this will fix some of the flow errors, but they're across all files.  They're all pretty mechanical to fix but will just take probably 2-3 hours to burn through them all, and I couldn't finish that yesterday morning and wanted to try to help unblock your other work first.

So I changed this PR to "work in progress."  Let's leave this open, and then revisit next week and either remove Flow altogether or finish this off.  If you have thoughts on that, or questions about how we might make that decision, feel free to comment here and we can talk more!

PR #345, opened 2018-03-09
Revert the Revert - New build process for Teacher Moments using create-react-app

kevinrobinson:
This reverts the revert in mit-teaching-systems-lab/threeflows#344, to ship https://github.com/mit-teaching-systems-lab/threeflows/pull/339.

But https://github.com/mit-teaching-systems-lab/threeflows/pull/346 is the missing piece to not bring the site down, so that should be merged into this PR first :)

PR #346, opened 2018-03-09
Update server code to read static files from create-react-app build

kevinrobinson:
This is a follow-on to https://github.com/mit-teaching-systems-lab/threeflows/pull/339.  This PR is intended to be merged into https://github.com/mit-teaching-systems-lab/threeflows/pull/344 (which reverts the revert, to deploy the new build process).

This changes the way the server code reads static assets like `index.html` and `.js` files in production, since we moved the UI code from `ui/` to `client/` and because the files that create-react-app builds are slightly different than the previous build process.


kevinrobinson:
@keving17 This is a PR into the double revert PR here: https://github.com/mit-teaching-systems-lab/threeflows/pull/345, not a PR into master.

PR #349, opened 2018-03-21
Feature/transcription

keving17:
Takes a wav file and transcribes it using IBM Watson. This is mostly a proof of concept.

I changed the Get Link button on the login page to try out the new HTTP server call.

Basically, I check the database for the transcript of a certain wav file. Rows in the evidence table have a `downloadUrl` field with a link to get the wav file in s3 - in this case, the `downloadUrl` was:
```
  audioFile = "blob:http://localhost:3000/a510a130-f17d-4331-90fe-1f2c8c0a2083";
```
I'm not entirely sure why the downloadUrl looks like that on my local database but that's fine.

If the transcript exists, I send a server response with a transcript field in the json body. Simply grabbing the transcript from the database takes about 30ms.
screen shot 2018-03-20 at 8 44 59 pm

If the transcript does not exist, I send an actual wav file
```
  audioFile = '/Users/keving17/Documents/Github/TSL/teacher-moments/server/20180213-185953-485_8d44f67e-d783-4597-aec2-d417325cce06.wav';
```
to IBM Watson's Speech to Text API and extract the full transcript from the response. I store this transcript and return it in a server response with the transcript in the body. This takes 15s.
screen shot 2018-03-20 at 8 51 58 pm

(The idea is for both images to look identical. One HTTP call took 500x longer to process though)

To get this fully working, I need to make sure I can get live data from our psql  database and wav files from our s3 locker. That way I can stop overwriting the `audioFile` variable and just send the `downloadUrl` in the request header.

For now, I'm using a hobby IBM account to transcribe things but the subscription plan looks pretty reasonable. I've run the actual service by YJ and she is fine having WWA pay for it if I use it. If this looks good so far, I'll spend the time to merge this into the researcher portal after the live data bit is finished!
screen shot 2018-03-20 at 9 01 05 pm


client/src/LoginPage.js
```diff
@@ -25,11 +24,13 @@ class LoginPage extends Component {
   }
 
   onSubmit(e) {
+    const audio = '/Users/keving17/Documents/Github/TSL/teacher-moments/server/20180213-185953-485_8d44f67e-d783-4597-aec2-d417325cce06.wav';
```
kevinrobinson: This is some good hacking to get a way to test this out! 👍  I'd suggest trying out new features from a clean scratch space (or at least a new component).  That usually makes it easier to unentangle things and get it into a state you can ship once you get the proof of concept working, but it's always hard to find the shortest path to hacking a prototype :)

If it's helpful at all, https://github.com/mit-teaching-systems-lab/transcriber/blob/master/transcribe.js is an example of how I approached that when I played with a similar thing for Google Cloud Speech API a while back.  At the time the quality wasn't good enough to be usable, but you can see it's just a few lines doing the proof-of-concept and so easier to develop and debug, and then once it's working you can figure out how to include it into the actual app.

server/speech.js
```diff
@@ -0,0 +1,119 @@
...
+// returns a promise resolved with the transcript
+function speechToText(pool, audioFile) {
+  //changing audioFile here to a wav file on my local disk to transcribe
+  //Should not need this line when hooked up to s3
+  audioFile = '/Users/keving17/Documents/Github/TSL/teacher-moments/server/20180213-185953-485_8d44f67e-d783-4597-aec2-d417325cce06.wav';
+  
+  var SpeechToTextV1 = require('watson-developer-cloud/speech-to-text/v1');
+  var fs = require('fs');
+
+  var watson = new SpeechToTextV1 ({
```
kevinrobinson: This is GitHub, so your credentials are now public :)  I'm sure this is fine for your hobby demo account, but in general you should store these kinds of things in environment variables.  Then this changes to `process.env.WATSON_USERNAME` and when you run the server locally you can do `WATSON_USERNAME=foo yarn start` or whatever.

server/speech.js
```diff
@@ -0,0 +1,119 @@
...
+            resolve(results);
+          });
+      }
+    });
+  });
+}
+
+// Updates database with transcript so we dont need to transcribe 
+// the same wav files multiple times
+function updateTranscript(pool, text, audioFile) {
```
kevinrobinson: Yeah this is awesome!  So in regards to my comment #3 above around how the engineering design works, the way you wrote this code in different functions is great.  You could shuffle it around and re-compose it into different designs, depending on the tradeoffs you want in terms of whether all past data is transcribed, how different engineering designs impact the researcher user experience, etc.

Right now, the work to do the transcription is on the "read path" - when the user asks for a response, the server goes and computes the transcription then, on-demand, if it's needed.  Another design could move this to the "write path," where whenever the server receives new audio evidence, it writes it to the database, and runs the transcription then.  Or another design is that could move this to an "offline job" that does the work in batches periodically for all the recent audio evidence.

Anyway, I think focusing on #1 transcription quality and #2 data sharing are best to start with, but just calling out that the way you wrote this proof-of-concept is great for then iterating on the final design that gets the tradeoffs you want for cost / user experience / etc.

kevinrobinson:
Also, this PR is a hard to review all at once, since it's on top of the other branch.  If that's important (which sometimes happens when there's lag between what you're doing and when the reviewing is responding), you can also create a PR that proposes merging your branch into another branch that's not master.  So here that might be a PR that merges `feature/transcription` into `feature/authentication`.  That way the PR diff is just the commits that are part of this new feature.  No need to change anything now, just sharing - it's always hard to figure out how to cut up PRs when there's lag in reviews.

keving17:
Since I ended up finishing this work on feature/watson-transcribe branch, I'll go ahead and close this PR.

PR #353, opened 2018-04-04
Feature/react virtualized

keving17:
This work was forked off of feature/authentication to upgrade the data table into a responsive table that could be sorted. Currently, I only show timestamps, email addresses, prompts and responses. Something to think about is how to control which email addresses are shown vs which are not since teacher educators may need to know the identity of users vs researchers who should not know. I'm defaulting to just showing the email addresses now since we'll need it for the WWA playtest. @kevinrobinson, what do you think?

Wider screen view of the data table
screen shot 2018-04-04 at 2 01 31 pm

narrow view of the data table
screen shot 2018-04-04 at 2 01 45 pm

sorting table by prompt
screen shot 2018-04-04 at 2 04 20 pm


keving17:
merging!

kevinrobinson:
@keving17 this looks awesome - way more usable and faster too, nice work! 👍

Issue #354, opened 2018-04-05
How does insecureStreamAudioFileFromS3 work??


keving17:
I'm trying to get the audio players in ResearcherDataPage.js to fetch audio files from s3 and allow users to play the audio files in the browser.
<img width="748" alt="screen shot 2018-04-05 at 12 46 23 am" src="https://user-images.githubusercontent.com/11862658/38347673-cb28ba4c-386a-11e8-9658-553529b9342e.png">

I'm looking at insecureStreamAudioFileFromS3 at [audio.js](https://github.com/mit-teaching-systems-lab/threeflows/blob/master/server/endpoints/audio.js) and I have no idea how it works. Are there any examples of code that use this function? I wasn't able to find anything to hint towards how to use this.

My intuition is that I should be able to either

1. download the wav file and keep it in the background as some resource - similar to how we had originally downloaded the actual wav files and referenced them locally in the data table

2. generate some link that just allows me to stream the wav file when I hit play

How I'm understanding the code at insecureStreamAudioFileFromS3 is that we are grabbing an object from s3, streaming the audio as s3 sends it over and piping into the response object. However, I cannot find anything in the response object. There is a 200 response code indicating that something was successfully done but I can't figure out what happened.

I've started a new branch feature/getAudio where I assumed insecureStreamAudioFileFromS3 returned some wav file I could use at the source for a html audio object. Basically the source for the audio players start out as a pending promise, and then when the promise is resolved, I go back and find the audio element and update its source to whatever result I get back - intuitively, this result is whatever was grabbed from s3. However, this turns out to just be an [object Response] so I think I'm back where I have no idea how to extract the streamed audio from the response that it was piped into. You can see this at commit https://github.com/mit-teaching-systems-lab/threeflows/commit/c99325ad2d48e25c73706a777e4e2303066748c1.

Since the common problem seems to be not knowing how the piping s3 audio into the result object works, I'm playing around with concatenating all of the buffered data that is streamed in streamed and creating my own blob. Unfortunately, this bit seems to be getting more complicated and I fear I'm digging too far down into this rabbit hole.

Any thoughts/insights on how exactly to use insecureStreamAudioFileFromS3 would be very much appreciated. 🏳️

keving17:
Hmmm, so I found a [resource](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) that seems to show how to extract the stream of data that was piped into the response.
[...]

PR #355, opened 2018-04-05
Feature/get audio

keving17:
Sets up ResearcherDataPage to grab audio wav files from s3. I copied the `insecureStreamAudioFileFromS3` function from `audio.js` into `audioEndpoint` so requests for audio are automatically granted as long as

1. they have a token to pass OnlyAllowResearchers

1. they have a id for the audio file they want taken from the dataEndpoint. This also means they already passed the checkAccess function which currently only gives access to my test data but in future work, will decide which researchers have access to which sets of data

screen shot 2018-04-05 at 2 05 24 am

I discussed how I translated the stream of audio from s3 into a wav blob at issue https://github.com/mit-teaching-systems-lab/threeflows/issues/354.

### DIFF
--- a/client/src/research/ResearcherDataPage.js
+++ b/client/src/research/ResearcherDataPage.js
@@ -211,23 +211,7 @@ class Analysis extends Component {
       method: 'GET'
     })
       .then(results => {
-        const reader = results.body.getReader();
-        const stream = new ReadableStream({
-          start(controller) {
-            function push() {
-              reader.read().then(({done, value}) => {
-                if (done) {
-                  controller.close();
-                  return;
-                }
-                controller.enqueue(value);
-                push();
-              });
-            }
-            push();
-          }
-        });
-        var response = new Response(stream, {headers: {"Content-Type": "audio/wav"}});
+        var response = new Response(results.body, {headers: {"Content-Type": "audio/wav"}});
         response.blob().then(function(myBlob) {
           var bloburl = URL.createObjectURL(myBlob);
           var elementTarget = document.getElementById(audioID.slice(0, -4));
--- a/client/src/research/ResearcherDataPage.js
+++ b/client/src/research/ResearcherDataPage.js
@@ -31,13 +31,7 @@ const styles = {
 };
 
 function getAudioUrl(row) {
-  console.log(row)
   return row.json.audioUrl || (row.json.audioResponse && row.json.audioResponse.audioUrl) || (row.json.uploadedUrl);
-}
-function rewriteAudioUrl(s3Folder, audioUrl) {
-  const slashIndex = audioUrl.lastIndexOf('/');
-  const filename = audioUrl.slice(slashIndex + 1);
-  return `${s3Folder}${filename}`;
 }
 
 function statsGroupedBy(allRows, keyName, keyFn) {
@@ -208,8 +202,7 @@ class Analysis extends Component {
   getAudio(audioID,elementID) {
     const token = this.state.token;
 
-    console.log('fetch:','/server/research/wav/'+audioID);
-    return fetch('/server/research/wav/'+audioID, {
+    fetch('/server/research/wav/'+audioID, {
       headers: {
         'Accept': 'application/json',
         'Content-Type': 'application/json',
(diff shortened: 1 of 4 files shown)


keving17:
selfie

client/src/research/ResearcherDataPage.js
```diff
@@ -193,6 +191,41 @@ class Analysis extends Component {
...
+      },
+      method: 'GET'
+    })
+      .then(results => {
+        var response = new Response(results.body, {headers: {"Content-Type": "audio/wav"}});
+        response.blob().then(function(myBlob) {
+          var bloburl = URL.createObjectURL(myBlob);
+          var elementTarget = document.getElementById(audioID.slice(0, -4));
+          if (elementTarget) {
+            document.getElementById(audioID.slice(0, -4)).src = bloburl;
```
kevinrobinson: As a rule of thumb, you shouldn't ever be calling `document.getElementById` in React components.  If you really really need to work directly with a DOM element in React, you can use `refs`. But usually it's better to call `setState` with what new data or information you got and then let `render` read that data and convert it to HTML.  Here, if you keep this approach to getting audio blobs, you would take the new blob reference, store it in the component's state, and then express what that should look like in HTML in `render` methods.

Separately, I think it's also worth exploring if you can directly render an `<audio />` tag with the URL set to the server URL for a particular `audioID`, or understanding where that's stuck.  That's what the private repo did previously, and that would let you remove all this code.  That may not work if it doesn't let you send the token in the header along with the request.  If that's the case another way to improve this would be to exchange the token with the server for a cookie on login, since I believe the browser will send cookies with `<audio />` requests.

Regardless this is all another reason to factor all this complexity out into its own `<AudioWithTranscription />` component or something like that, and remove the DOM manipulation code (related to https://github.com/mit-teaching-systems-lab/threeflows/pull/348/files#r180895553).

server/getAudio.js
```diff
@@ -0,0 +1,42 @@
...
+function getAudioKey(request, id) {
+  const deploymentFolder = [process.env.NODE_ENV, request.headers.host].join('_');
+  return [
+    deploymentFolder,
+    'wav-responses',
+    `${id}.wav`
+  ].join('/');
+}
+
+function insecureStreamAudioFileFromS3(params, request, response) {
```
kevinrobinson: @keving17 Import this rather than copying and pasting?

server/index.js
```diff
@@ -214,12 +215,14 @@ if (process.env.ENABLE_RESEARCHER_ACCESS && process.env.ENABLE_RESEARCHER_ACCESS
...
   // Endpoints for authenticated researchers to access data
   app.get('/server/research/data', [limiter, onlyAllowResearchers.bind(null, pool)], dataEndpoint.bind(null, pool));
+  app.get('/server/research/wav/(:id).wav', [limiter, onlyAllowResearchers.bind(null, pool)], audioEndpoint.bind(null, pool, config.s3));
 }
 
 // Serve any static files.
 // Route other requests return the React app, so it can handle routing.
 app.use(express.static(path.resolve(__dirname, '../client/build')));
 app.get('*', (request, response) => {
+  console.log('caught');
```
kevinrobinson: leftover

scripts/updateConsent.js
```diff
@@ -18,14 +18,15 @@ function batchPromises(batchSize, emails, fn){
 
 function updateConsent(database){
   //Still need to grab emails from a sensitive data file
-  const emails = JSON.parse(fs.readFileSync('./tmp/consented-latest.json'))
+  // const emails = JSON.parse(fs.readFileSync('./tmp/consented-latest.json'))
+  const emails = JSON.parse(fs.readFileSync('./tmp/test.json'))
```
kevinrobinson: This seems like the wrong filename :)

client/src/research/ResearcherDataPage.js
keving17: Oh yes. This was my last ditch effort to get this done before the Friday playtest. I was planning on the audioplayers into their own component. I made issue https://github.com/mit-teaching-systems-lab/threeflows/issues/362 to make sure I do this

server/getAudio.js
keving17: Oh yeah, I originally copied this function over because I couldnt figure out how to use it and ended up changing it a bunch. I had a lot of problems with this initially as seen in https://github.com/mit-teaching-systems-lab/threeflows/issues/354.

Since I ended up just using the original function, I'll import this instead. Should I reopen this branch to make this change or just do it in a patch on master?

server/index.js
keving17: whoops. Again, should I reopen this branch to make this change or just do it in a patch on master?

scripts/updateConsent.js
keving17: Oh yes, this needs to be changed. Again, reopen this branch or just do a patch?

I also might need some help figuring out how to run this script on heroku 😭

server/getAudio.js
kevinrobinson: @keving17 sounds good!  for new changes, make a new branch for each set of changes.

scripts/updateConsent.js
keving17: Hmmm. actually, it seems like I fixed this already in commit https://github.com/mit-teaching-systems-lab/threeflows/commit/6b1b2762302b8d6c601d6e83798009eda79e4744

Issue #350, opened 2018-03-22
Fixup README with db-create tasks, remove database table SQL from README


kevinrobinson:
Hey @keving17!  I was trying to run your awesome work in https://github.com/mit-teaching-systems-lab/threeflows/pull/348 locally, but noticed that the README is all out of date.  I think I remember you updating this before, but maybe it got lost in some git-fu somewhere.  Can you update [this part](https://github.com/mit-teaching-systems-lab/threeflows#local-software-development) of the README instructions?

keving17:
https://github.com/mit-teaching-systems-lab/threeflows/pull/356#pullrequestreview-111400062
Add Watson information to README

PR #356, opened 2018-04-06
Feature/watson transcribe

keving17:
Using IBM Watson Speech-To-Text to transcribe audio files to text for analyses tools.
Will need to insert watson credentials into production environment for this to work

screen shot 2018-04-06 at 2 51 45 am

### DIFF
--- a/client/src/research/Transcribe.js
+++ b/client/src/research/Transcribe.js
@@ -1,5 +1,5 @@
 function requestTranscript(token,audioID) {
-  return fetch('/server/research/transcribe/'+audioID, {
+  return fetch('/server/research/transcribe/'+audioID+'.wav', {
     headers: {
       'Accept': 'application/json',
       'Content-Type': 'application/json',
--- a/client/src/research/Transcribe.js
+++ b/client/src/research/Transcribe.js
@@ -1,0 +1,20 @@
+function requestTranscript(token,audioID) {
+  return fetch('/server/research/transcribe/'+audioID, {
+    headers: {
+      'Accept': 'application/json',
+      'Content-Type': 'application/json',
+      'x-teachermoments-token': token,
+    },
+    method: 'POST'
+  })
+    .then(result => {
+      return result.json();
+    })
+    .catch(err => {
+      console.log('Failed to request transcript');
+    });
+}
+
+module.exports = {
+  requestTranscript
+};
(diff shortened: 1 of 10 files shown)


keving17:
selfie

server/speech.js
```diff
@@ -0,0 +1,106 @@
+// This endpoint handles server calls to get transcript of an audio file
```
kevinrobinson: @keving17 This is really well done with the way you've factored this into the different pieces! 👍

server/speech.js
```diff
@@ -0,0 +1,106 @@
...
+  const values = [audioID];
+  const sql = `
+    SELECT transcript
+    FROM transcripts
+    WHERE 1=1
+      AND audio_id = $1`;
+
+  pool.query(sql,values)
+    .then(results => {
+      if ((results.rowCount >= 1) && (results.rows[0].transcript !== "")){
```
kevinrobinson: Would adding `LIMIT 1` to the SQL simplify these checks?

server/watson_client.js
```diff
@@ -0,0 +1,8 @@
+var fs = require('fs');
+
+module.exports = function createWatsonClient() {
+  const watsonConfig = (process.env.NODE_ENV === 'development')
+    ? JSON.parse(fs.readFileSync('./tmp/watson_config.json'))
+    : JSON.parse(process.env.WATSON_CONFIG_JSON);
```
kevinrobinson: This probably needs a line in the README.

client/src/research/Transcribe.js
```diff
@@ -0,0 +1,20 @@
...
+      'Content-Type': 'application/json',
+      'x-teachermoments-token': token,
+    },
+    method: 'POST'
+  })
+    .then(result => {
+      return result.json();
+    })
+    .catch(err => {
+      console.log('Failed to request transcript');
```
kevinrobinson: should this return a failure status for the HTTP request?

server/speech.js
keving17: So I think that since I use the 0th item in rows anyways, it is probably fine to have `LIMIT 1`. 

Do you think it would be better to use `LIMIT 1` and just have the if statement be 
```
if ((results.rowCount === 1) && (results.rows[0].transcript !== "")){
```

I'm not sure if I'm missing the point.

server/watson_client.js
keving17: yep! I added this to issue https://github.com/mit-teaching-systems-lab/threeflows/issues/350

client/src/research/Transcribe.js
keving17: hmmmm. So I think I generally have the return statuses be set on the server side. In this case, in getAudio.js, the insecureStreamAudioFileFromS3 returns a 200 by default if it can find the audio file and returns a 503 if it can't get an audio file from s3.

PR #360, opened 2018-04-19
fix email mustache to include login link

keving17:
Addresses Issue https://github.com/mit-teaching-systems-lab/threeflows/issues/357.
When I migrated this code from the swipe right project, I changed how some things were named and broke this bit on the emailing functionality. This fixes that.

screen shot 2018-04-19 at 3 30 17 pm

PR #363, opened 2018-04-23
Feature/access

keving17:
Gives access to data based on a new access db that links emails to location urls

Edit: Turns out this includes the commit for pull request https://github.com/mit-teaching-systems-lab/threeflows/pull/360 for fixing the login link. Not sure what to do about that part.

### DIFF
--- a/server/database.js
+++ b/server/database.js
@@ -6,11 +6,16 @@ const {getDomain} = require('./domain.js');
 //Returns 200 unless there was an error with the database query
 function dataEndpoint(pool, request, response) {
   const token = request.headers['x-teachermoments-token'];
+  const url = request.headers['x-teachermoments-location'].split(',');
+  console.log('url',url)
+
   const domain = getDomain(request);
-  const location = `${domain}`+process.env.ANALYSIS_LOCATION;
+  const location = `${domain}`+url[0];
+  console.log('location',location)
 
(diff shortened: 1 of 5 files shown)


keving17:
Checked that everything works fine locally. Going to deploy this now so Meredith can look at her data tomorrow morning

server/database.js
```diff
@@ -32,16 +34,52 @@ function dataEndpoint(pool, request, response) {
     });
 }
 
-// Spoofing the part where I confirm a token is allowed to view data from 
-// a specific location. Location is currently hard coded so this is fine.
+// It seems more natural to store email/url pairs in access db.
+// Should I be using token to find the corresponding email and 
+// use the email to check for access?
+// I'm not sure how to store token/url pairs.
```
kevinrobinson: @keving17 It looks like you figured this out and got this working, right?

Checking my understanding: the UI passes `location`, the server looks up to see if the token is valid, then looks up to see if that email is valid for that location, then sends back data for that location.  Is that right?

keving17: yep that's right! since the UI only send the token, we have to do an extra query to look up what email corresponds to the given token but then we can use that email to see if it is authorized to grab data from that location.

kevinrobinson: 👍

PR #365, opened 2018-04-25
change script to read from stdin

keving17:
changes updateConsent.js to read from standard input


keving17:
selfie. did this with @kevinrobinson

Issue #367, opened 2018-04-26
emails should be case-insensitive


keving17:
There were some queries that failed to find users if they did not use all lowercase when they entered email into threeflows. Just add the LOWER function to psql queries

PR #371, opened 2018-06-01
ignores case sensitivity in emails when grabbing data

keving17:
ignores case sensitivity when when matching emails users put into teacher moments against their consent form. Resolves https://github.com/mit-teaching-systems-lab/threeflows/issues/367


keving17:
selfie

PR #372, opened 2018-06-01
Patch/readme

keving17:
This patch updates the README with all of the new scripts introduced with the new researcher portal. It address issue https://github.com/mit-teaching-systems-lab/threeflows/issues/350.

Notable updates include:
- new script to seed database
- small section on the Watson API and necessary credentials
- Viewing data section to introduce login portal
- jest section to replace mocha
- Maintaining Researcher Portal section to explain how to update consented users on live site.

[View changes here](https://github.com/mit-teaching-systems-lab/threeflows/blob/patch/readme/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ $ yarn db-create-dev
 ```
 - For storing audio files, you'll need keys authorized for a development S3 bucket.
 - For transcribing audio files, you'll need keys authorized for accessing the Watson Speech to Text API. Details on [how to use the API](https://www.ibm.com/watson/developercloud/speech-to-text/api/v1/) can be found on their website.
+
 Ask someone for these keys.
 
 #### Viewing data
--- a/README.md
+++ b/README.md
@@ -149,11 +149,11 @@ To update Teacher Moments on who has consented to share their data, you need to:
 ```
 $ yarn run prep-consent
 ```
-This should generate 2 files: `consented-latest.json` and `consented-YYYY-MM-DD.json`.
+This should generate 1 file in the `threeflows/tmp` folder: `consented-latest.json`.
 
-3. Run the update-consent script in heroku to update the live `consented_email` database using
+3. Run the update-consent script in heroku to update the live `consented_email` database and clean sensitive information from your codebase using
 ```
-$ cat ./tmp/consented-latest.json | heroku run --no-tty yarn run update-consent
+$ cat ./tmp/consented-latest.json | heroku run --no-tty yarn run update-consent | rm ./tmp/consented-latest.json ./tmp/consented-latest-raw.csv
 ```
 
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -162,6 +130,33 @@ Other Heroku docs:
...
+All data collected by Teacher Moments is protected to ensure the privacy of participants. In order for a researcher or TE to view the data collected on a specific participant, the participant must fill out a [TSL consent form.](https://tsl.mit.edu/consent)
+
+To update Teacher Moments on who has consented to share their data, you need to:
+1. Download a CSV version of the consent spreadsheet. Search for `User Testing Consent Form (Responses)` in the TSL team Google Drive. Rename the file to `consented-latest-raw.csv` and save this file in the `threeflows/tmp` folder.
+
+2. Run the prep-consent script to process the CSV file using
+```
+$ yarn run prep-consent
+```
+This should generate 2 files: `consented-latest.json` and `consented-YYYY-MM-DD.json`.
```
kevinrobinson: These are will be written into the `tmp` folder, right?

README.md
```diff
@@ -162,6 +130,33 @@ Other Heroku docs:
...
+
+3. Run the update-consent script in heroku to update the live `consented_email` database using
+```
+$ cat ./tmp/consented-latest.json | heroku run --no-tty yarn run update-consent
+```
+
+
```
kevinrobinson: Do you think it's worth adding a note here that these are sensitive files and you should delete the `tmp` folder after doing this operation?

README.md
keving17: yes these are written to the tmp folder

README.md
keving17: so the thing is, we also use the tmp folder to keep the credentials for the s3 and watson speech to text APIs. I haven't thought too much about security in this case since everything in the tmp folder is in the .gitignore file. 

Do you think we should add a note? I don't think it would hurt. But I also thought that the old script made a copy of the latest-consent.json with the date for record keeping. If we want people to delete the csv and json files after this operation, I should probably take out the `consented-YYYY-MM-DD.json` from the script.

kevinrobinson: Maybe more targeted then, like `cat ./tmp/consented-latest.json | heroku run --no-tty yarn run update-consent | rm ./tmp/consented-latest.json`?

The change to remove the `consented-YYYY-MM-DD.json` file sounds good, that's leftover from how the private repo did this in the file system, and you've improved that to a better system in the database now!  So it'd be great to make these more temporary files that are just there while you're updating consent, and gone after.

re: security, the consent files are more sensitive than any of those other keys.  If we leaked one of the keys, we can rotate it immediately and then it's useless and can stay public forever with no ill effects.  If we leak people's email addresses, there's nothing we can do to recover from that being public.  So while access keys and credentials are very sensitive, the consent file is even more sensitive.

keving17: that makes sense. Fixed that stuff!

kevinrobinson:
@keving17 mergin'! :)

Issue #358, opened 2018-04-06
Transcripts seem to cut off early


keving17:
The only example I have so far involves sensitive user data. I'll try to come up with an example to screenshot later

PR #374, opened 2018-06-11
Patch/short transcript

keving17:
Resolves Issue https://github.com/mit-teaching-systems-lab/threeflows/issues/358 where transcripts would cut off early if users had an extended pause in their audio responses.

screen shot 2018-06-10 at 8 16 52 pm


keving17:
selfie

Issue #362, opened 2018-04-23
Move audio players into their own react component


keving17:
Instead of finding elements and updating the source of <audio> tags, move all of this into its own component so react can take care of updating the component when the audio is pulled from s3. See https://github.com/mit-teaching-systems-lab/threeflows/pull/355#pullrequestreview-111399693

Issue #361, opened 2018-04-23
Refactor audio and transcript loading into separate components


keving17:
https://github.com/mit-teaching-systems-lab/threeflows/pull/348#pullrequestreview-111399627

PR #375, opened 2018-06-11
Patch/audio component

keving17:
Addresses Issues https://github.com/mit-teaching-systems-lab/threeflows/issues/362 and https://github.com/mit-teaching-systems-lab/threeflows/issues/361 about moving audio players and transcripts to their own separate component rather than finding elements by ID and updating them as data was retrieved from s3 and Watson API. 

This adds two new components to `DynamicTable.js`: `AudioPlayerComponent` and `TranscriptComponent` which update the table as data is returned.

### DIFF
--- a/client/src/research/DynamicTable.js
+++ b/client/src/research/DynamicTable.js
@@ -162,6 +162,26 @@ class DynamicHeightTableColumn extends React.PureComponent {
     );
   };
 
+  _getAudioAndTranscript = ({ dataKey , rowData }) => {
+    return (
+      rowData.json[dataKey] || (this._getAudioUrl(rowData) && 
+      <div>
+        <AudioPlayerComponent audioID={this._getAudioID(this._getAudioUrl(rowData))} token={this.props.token} /> 
+        <TranscriptComponent audioID={this._getAudioID(this._getAudioUrl(rowData))} token={this.props.token}/> 
+      </div>)
+    );
+  };
+
+  _getPrompt = ({ dataKey , rowData }) => {
+    return (
+      rowData.json.question.text || 
+      <div>
+        <span>Teacher Moments Scene: </span> 
+        <a href={"https://youtu.be/"+rowData.json.question.youTubeId}>https://youtu.be/{rowData.json.question.youTubeId}</a>
+      </div>
+    );
+  };
+
   render() {
     const width = this.props.width;
 
@@ -223,7 +243,7 @@ class DynamicHeightTableColumn extends React.PureComponent {
           dataKey="text"
           label="Prompt"
           disableSort = {false}
-          cellDataGetter={({ dataKey , rowData }) => rowData.json.question.text || <div><span>Teacher Moments Scene: </span> <a href={"https://youtu.be/"+rowData.json.question.youTubeId}>https://youtu.be/{rowData.json.question.youTubeId}</a></div>}
+          cellDataGetter={this._getPrompt}
           cellRenderer= {this._cellRenderer}
           headerRenderer={this._headerRenderer}
           width={width}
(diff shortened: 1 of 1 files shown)


keving17:
I tested this locally. Everything works fine except for a minor visual bug when the table first loads up. That can be fixed by adjusting the window size slightly. More details in Issue https://github.com/mit-teaching-systems-lab/threeflows/issues/376

client/src/research/DynamicTable.js
```diff
@@ -344,12 +230,105 @@ export default class DynamicHeightTableColumn extends React.PureComponent {
...
+      })
+      .catch(err => {
+        console.log('failure in transcription');
+      });
+  }
+
+  render() {
+    return (
+      <div id={this.state.audioID+"-transcript"}>Transcript: "{this.state.transcript}"</div>
+      // <div id={audioID+"-transcript"}>Transcript: "{results.transcript}"</div>
```
kevinrobinson: leftover comment

client/src/research/DynamicTable.js
```diff
@@ -344,12 +230,105 @@ export default class DynamicHeightTableColumn extends React.PureComponent {
           dataKey="responseText"
           label="Response"
           disableSort = {false}
-          cellDataGetter={({ dataKey , rowData }) => rowData.json[dataKey] || (this._getAudioUrl(rowData) && <div>{this._loadAudioPlayer(this._getAudioID(this._getAudioUrl(rowData)))} {this._loadTranscript(this._getAudioID(this._getAudioUrl(rowData)))} </div>)}
-          cellRenderer= {this._wrappingCellRendererAgain}
+          cellDataGetter={({ dataKey , rowData }) => rowData.json[dataKey] || (this._getAudioUrl(rowData) && <div><AudioPlayerComponent audioID={this._getAudioID(this._getAudioUrl(rowData))} token={this.props.token} /> <TranscriptComponent audioID={this._getAudioID(this._getAudioUrl(rowData))} token={this.props.token}/> </div>)}
```
kevinrobinson: This is a lot in one line :)  maybe make a method that `cellDataGetter` can call?

client/src/research/ResearcherDataPage.js
```diff
@@ -281,156 +197,6 @@ class Analysis extends Component {
...
-          </tbody>
-        </table>
-      </div>
-    );
-  }
-
-  // For looking at raw events and sessions.  Hover to reveal more sensitive data.
-  renderEventsTable(json) {
-    const {analysisKey} = this.props;
-    if (analysisKey === 'HMTCA') return this.renderHmtcaTable(json);
```
kevinrobinson: This looks like it's simplifying, but also that it's removing some code that is used to analyze other data besides the scenarios you've been focused on.  Can you say more about how removing this code impacts viewing data for other sessions or scenarios?

If removing this would break those, let's leave this code and add a branching to use the new awesome view where it works and for older analyses use this old code.  Or if it didn't really work before this change to look at the older analyses, then I think it's fine to remove this and just add a comment linking back to an older commit to point future folks in a helpful direction.

client/src/research/DynamicTable.js
```diff
@@ -344,12 +230,105 @@ export default class DynamicHeightTableColumn extends React.PureComponent {
...
           width={width}
         />
       </Table>
     );
   }
-}
\ No newline at end of file
+}
+
+class AudioPlayerComponent extends React.Component {
```
kevinrobinson: This looks great.  Can you move `AudioPlayerComponent` and `TranscriptComponent` into their own files?  They're now nicely self-contained which simplifies the table code quite a bit but also could be re-used if other folks are making other kinds of analysis.  In the same file here they're a bit hidden away.

keving17: done

client/src/research/ResearcherDataPage.js
keving17: So those stopped working once I stopped using the renderEventsTable() function and switched to using react virtualized. Are you saying to leave a comment in the actual code about this?

client/src/research/DynamicTable.js
keving17: done!

keving17:
So overall, this PR should not change any of the behavior since commit https://github.com/mit-teaching-systems-lab/threeflows/commit/a3ffebd00250cd5600376c648edb91e3a10fcc3a where I fixed up some problems with using react-virtualized. I only create and move the AudioPlayerComponent and TranscriptComponent outside of the constructor for the data table and clean up some leftover code from when I switched over to react-virtualized.

keving17:
If things look ok to you @kevinrobinson , I can merge this.

client/src/research/ResearcherDataPage.js
kevinrobinson: Yep!  I think it's fine to break these older scenarios to make the awesome improvements you've made.  I'd suggest just adding comment that says something like "for analysis of other scenarios, look at (this URL in the git history)" as a helpful pointer for future folks.

kevinrobinson:
@keving17 sounds great, 🚢!

PR #378, opened 2018-09-19
Fix migration bug that dropped Rollbar

kevinrobinson:
This must have gotten dropped in migrating the build process, so let's add it back to get more visibility into errors, especially with more new folks trying these out in more self-serve ways.

Also address linter warning on master.


kevinrobinson:
selfie cc @TheGarron

Issue #377, opened 2018-09-17
Audio Recorder fails to load and screen, becomes blank white breaking the simulation


TheGarron:
Reported by Sheila Conway (partner researcher) via email
"After the first video the screen goes completely white.  I do not get a microphone use prompt that others get and can't see video any longer. My only option is to hit back arrow on search bar and it bring me back to start of program (list of demo projects).
I tried to use this video on several other occasions with same problem."

TheGarron:
After investigating a reproduction of this issue the behavior can be reproduced when the browser does not have access to the microphone. As this setting varies based on device/browser the reproduction can be achieved via browsers not used in development such as Safari on the desktop. Similarly when using an iPhone with the default browser of Safari the same issue can be reproduced.

This was reported on the Sheila Turner simulation which can be accessed on the teacher moments demos page by clicking on the link "Parent doesn't value school" which redirects to the following URL: http://localhost:3000/teachermoments/turner?fromdemos

This is will be marked as won't resolve as this is a browser configuration problem. I provided Sheila the link to Apple Support (https://support.apple.com/guide/safari/websites-preferences-ibrwe2159f50/mac) which covers changing the setting to allow the browser to have access to the microphone. I also recommended using chrome as the browser for the study as it is supported browser for this application. Chrome prompts the user to allow access to the microphone while Safari loaded in the iPhone does not appear to prompt for access meaning access must be granted ahead of using the application.

kevinrobinson:
@TheGarron nice!  To clarify, does this impact iOS Safari too?

I can't reproduce the problem at https://teachermoments.teachingsystemslab.org/teachermoments/turner on my iPhone 6 running iOS 11.  I don't see a prompt about audio recording and it falls back to text responses (calling into [AudioCapture.isAudioSupported](https://github.com/mit-teaching-systems-lab/threeflows/blob/master/client/src/components/audio_capture.jsx#L24)) rather than trying audio and breaking.

I was going to link to the relevant Rollbar issues, but in looking at this I saw we haven't got a Rollbar error since March, which seemed weird (more in a moment).  But here's a stack trace from 3/5/18 that I think is from someone experiencing what Sheila experienced:
```
TypeError: null is not an object (evaluating 'this.recordingStream.getTracks')
  File "https://teachermoments.teachingsystemslab.org/bundle.js", line 37, in value
  File "https://teachermoments.teachingsystemslab.org/bundle.js", line 37, in stopRecording
[...]

PR #380, opened 2018-09-20
Wrap isAudioSupport in try/catch

kevinrobinson:
Workaround for iOS Chrome 68 so the fall back to text works, rather than breaking.  See also https://github.com/mit-teaching-systems-lab/threeflows/issues/377#issuecomment-422954152 and  https://rollbar.com/threeflows/threeflows/items/147/#graphs

### DIFF
--- a/client/src/components/audio_capture.jsx
+++ b/client/src/components/audio_capture.jsx
@@ -22,12 +22,18 @@ export default class extends React.Component {
   };
 
   static isAudioSupported() {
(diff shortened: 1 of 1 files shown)


kevinrobinson:
@TheGarron I requested a review from you on this in GitHub, so you should get a notification on that, and then can comment and approve/suggest changes/reject, so you can see how that flow works.

PR #382, opened 2019-01-30
Remove QuestionsPage.test, which is causing HTTP requests and failing the build

kevinrobinson:
This is addressing the issue in https://travis-ci.org/mit-teaching-systems-lab/threeflows/builds/486552366?utm_source=github_status&utm_medium=notification and earlier builds.

This code hasn't been used by anyone in probably two years and it's blocking deploys, so removing the test is better than trying to dig in further today.


kevinrobinson:
@TheGarron To fix the immediate issue, this PR removes the test that is causing the issue.  It's from work an awesome UROP did in summer 2016 to allow end-user authoring, but no one is actively using it.

Separately, I rewrote the history for master to fix the issues with the commit history, which will impact both you and Natalie.  I can show you how to resolve that on your computer in person.

Backing up, the test shouldn't pass locally and fail in CI.  I think this is from the differences in how "watch mode" and "CI mode" work in Jest.  The command line options that newer versions support don't work in the current setup, so the way I could reproduce this was by adding `CI=true` in front of the `jest` script in `client/package.json`.  Then I started with the subset of files shown in https://travis-ci.org/mit-teaching-systems-lab/threeflows/builds/486552366?utm_source=github_status&utm_medium=notification and narrowed it down to three files where the issue was, and then stopped looking further :)

kevinrobinson:
mergin'

kevinrobinson:
@TheGarron Locally, to sync back up with master, first make sure you are on a clean branch with no changes when you run `git status`.

Then, run:

```
$ git fetch origin # get the latest changes from github
$ git checkout origin/master # checkout to github's master
$ git branch -f master origin/master # forcibly update your local master to match github
$ git checkout master checkout # checkout to your local master, which should now match github
```

Then you can check the tests pass.  If they don't, ping me!

PR #426, opened 2019-05-13
Update docs for deploying a fork

kevinrobinson:
Improving these notes with @TheGarron!


kevinrobinson:
@TheGarron this has notes about S3 too now.

